PR #3, opened 2016-01-08
Add List koans

Maikon:
@heruku 

Initial draft for lists koans. Suggestions and ideas for names, additional koans, modification of existing or anything else, are welcomed.

### DIFF
--- a/lib/koans/lists.ex
+++ b/lib/koans/lists.ex
@@ -1,7 +1,68 @@
 defmodule Lists do
   use Koans
 
-  koan "We shall contemplate truth by testing reality, via equality" do
-    assert List.first([1, 2, 3]) == 2
+  koan "We can see what is ahead" do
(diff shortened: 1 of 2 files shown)


lib/koans/lists.ex
```diff
@@ -1,7 +1,103 @@
...
+
+  koan "Order is bound by nature's laws" do
+    result_one = List.insert_at([1, 2, 3], 10, 4)
+    result_two = List.insert_at([1, 2, 3], -1, 4)
+
+    assert result_one == [1, 2, 3, 4]
+    assert result_two == [1, 2, 3, 4]
+  end
+
+  koan "Updating an existing element can be more complex" do
```
heruku: I was thinking that maybe functions should come after lists. In that case, this kata could be confusing to users :)

PR #14, opened 2016-01-08
Add support for blanks in assertions

heruku:
This is done by wrapping the ExUnit assertions that we use. Instead of calling ExUnit straight, the `BlankAssertions` module intercepts calls and does one of two things:

  * If the assertion contains a blank, it is failed immediately. ExUnit is never called and the failing expression is printed
  * If the assertions does not contain a blank, just delegates to ExUnit and lets it handle the assertion.


Maikon:
:+1: :100:

heruku:
This closes #14

PR #16, opened 2016-01-09
WIP: Start work maps

felipesere:
I want to start a discussion on what methods should be taught in the koans.
Once that is settled I will re-rewrite the examples to be more realistic, probably working on a map that represents a person.

### DIFF
--- a/lib/koans/03_maps.ex
+++ b/lib/koans/03_maps.ex
@@ -1,0 +1,59 @@
+defmodule Maps do
+  use Koans
+
+  koan "Maps have keys and values" do
+    assert %{ :the_key => "some value" } == %{ :the_key => "some value"}
+  end
+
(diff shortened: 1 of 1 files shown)


lib/koans/03_maps.ex
```diff
@@ -0,0 +1,59 @@
...
+
+  koan "Has keys" do
+    assert Map.keys(%{ :a => "A", :b => "B" }) == [:a, :b]
+  end
+
+  koan "Has values" do
+    assert Map.values(%{ :a => "A", :b => "B" }) == ["A", "B"]
+  end
+
+  koan "Can read values based on a key" do
```
heruku: I would split the first two into separate koans and possibly leave the `fetch!` version out for now.

PR #17, opened 2016-01-13
Intercept and print compile errors in a nicer way.

### LINKED ISSUE #4: Handle compile errors more gracefully
When the user saves a file with a compile error, the auto-runner does the right thing and keeps running, but the error is a bit scary:

screen shot 2016-01-08 at 15 54 37

Most of this is irrelevant to the user and we should be able to get rid of it.

ukutaht:
This PR changes the error formatting for compiler errors:
  * Only print the first two lines of the stacktrace to hide the runner internals from the trace.
  * It doesn't print the message about workers shutting down. That message is not useful to users.

Closes #4

### DIFF
--- a/lib/display.ex
+++ b/lib/display.ex
@@ -24,6 +24,11 @@ defmodule Display do
     "Assertion failed in #{source_file(module)}:#{line_number(expr)}"
   end
 
+  def format_compile_error(error) do
+    trace = System.stacktrace |> Enum.take(2)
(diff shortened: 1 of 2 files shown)


felipesere:
+1 LGTM.

ukutaht:
We'll have to make sure that the stack trace is sufficient for the function koans. Merging for now though.

PR #28, opened 2016-03-06
Basic Koans about structs

felipesere:


lib/koans/08_structs.ex
```diff
@@ -0,0 +1,59 @@
+defmodule Structs do
+  use Koans
+
+  defmodule Person do
+    defstruct [:name, :age ]
+  end
+
+  koan "structs are defined and named after a module" do
+    assert %Person{}
```
ukutaht: I understand the intention here but where would we leave a blank here? `assert __`  feels a bit strange to me

lib/watcher.ex
```diff
@@ -4,7 +4,7 @@ defmodule Watcher do
   def callback(file, events)  do
     if Enum.member?(events, :modified) do
       try do
-        [{mod, _}] = Code.load_file(file)
+        [{mod, _} | _] = Code.load_file(file)
```
ukutaht: What does this change?

felipesere: Structs are defined within a module, when I add a 
```
defmodule MyStruct do
  defstruct ...
defmodule
``` 
within a koan I get more than a single module. 
The previous version forced a single element in the list and failed the match.

lib/koans/08_structs.ex
felipesere: Yeah, I wanted some way to show how to create structs without going to deep. I can remove that koan.

PR #29, opened 2016-03-06
Koans about processes.

felipesere:
~~Includes fairly clunky infrastructure to show the right line number when a blank assertion fails.~~

### DIFF
--- a/lib/display.ex
+++ b/lib/display.ex
@@ -24,7 +24,8 @@ defmodule Display do
   end
 
   def display_failed_assertion(module, expr) do
-    "Assertion failed in #{source_file(module)}:#{line_number(expr)}"
+    file = source_file(module)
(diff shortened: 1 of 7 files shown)


lib/display.ex
```diff
@@ -31,10 +32,39 @@ defmodule Display do
     IO.puts(format_red(Exception.format(:error, error, trace)))
   end
 
-  defp line_number({_, [line: line], _}) do
-    line
+  defp line_number({_, [line: line], _}, in: _), do: line
+  defp line_number(expr, in: file) do
```
ukutaht: What was the problem here? Was it reporting the wrong line numbers?

felipesere: When doing `assert_receive <expr>` all I get passed in as the first argument to `line_number` is literally `<expr>`. I don't know if I have an error before that gets even called?

PR #32, opened 2016-03-07
Start basic pattern matching koans

ukutaht:
review @felipesere @Maikon

### DIFF
--- a/lib/koans/09_pattern_matching.ex
+++ b/lib/koans/09_pattern_matching.ex
@@ -59,11 +59,11 @@ defmodule PatternMatching do
   end
 
   koan "errors are shaped differently than sucessful results" do
-    result = case Integer.parse("obviously not a number") do
-      {parsed, _} -> parsed
-      :error -> 0
+    result = case Map.fetch(%{}, :obviously_not_a_key) do
+      {:ok, val} -> val
+      :error -> "not present"
     end
 
-    assert result == 0
+    assert result == "not present"
   end
 end
--- a/lib/koans/09_pattern_matching.ex
+++ b/lib/koans/09_pattern_matching.ex
@@ -6,9 +6,10 @@ defmodule PatternMatching do
   end
 
   koan "patterns can be used to pull things apart" do
-    [head | _tail] = [1,2,3,4]
+    [head | tail] = [1,2,3,4]
 
     assert head == 1
+    assert tail == [2,3,4]
   end
 
   koan "or put them back together" do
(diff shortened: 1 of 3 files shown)


lib/koans/09_pattern_matching.ex
```diff
@@ -0,0 +1,62 @@
...
+  use Koans
+
+  koan "one matches one" do
+    assert match?(1, 1)
+  end
+
+  koan "patterns can be used to pull things apart" do
+    [head | _tail] = [1,2,3,4]
+
+    assert head == 1
```
felipesere: I'd keep `tail` and also assert on it. 
Then introduce `_something` as away to explicitly ignore things.

lib/koans/09_pattern_matching.ex
```diff
@@ -0,0 +1,62 @@
...
+  koan "the pattern can make assertions about what it expects" do
+    the_list = [1, 2, 3]
+
+    assert match?([1, _, _], the_list)
+    refute match?([2, _, _], the_list)
+  end
+
+  def make_noise(%{type: "cat"}), do: "Meow"
+  def make_noise(%{type: "dog"}), do: "Woof"
+  def make_noise(_), do: "Eh?"
```
felipesere: same here: `_anything`?

lib/koans/09_pattern_matching.ex
```diff
@@ -0,0 +1,62 @@
...
+    snake = %{type: "snake", legs: 0, age: 20, color: "black"}
+
+    assert make_noise(dog) == "Woof"
+    assert make_noise(cat) == "Meow"
+    assert make_noise(snake) == "Eh?"
+  end
+
+  koan "errors are shaped differently than sucessful results" do
+    result = case Integer.parse("obviously not a number") do
+      {parsed, _} -> parsed
```
felipesere: This should probably be `{:ok, parsed}`

ukutaht: `Integer.parse` is actually weird
screen shot 2016-03-07 at 11 02 06


Maybe it's not the best function to show off error handling because you're right, `{:ok, result}` is way more common. Any ideas for what standard library function we could use here to demonstrate patterns in error handling?

felipesere: `{:error} == Enum.fetch([1,2,3], 50)`
vs
`{:ok, x} == Enum.fetch([7,8,9],2)`

felipesere: Or we can just call our own function...

felipesere:
LGTM

PR #37, opened 2016-03-08
Squash compiler warnings.

felipesere:


lib/koans/07_arithmetic.ex
```diff
@@ -53,8 +53,9 @@ defmodule Arithmetic do
   end
 
   koan "it can not divide by zero" do
```
ukutaht: I think arithmetic is a pretty basic koan so it may be too early to introduce anonymous function syntax here. I didn't find a way to suppress this warning easily either. What do you think about just deleting this for now?

felipesere: Will do.

PR #40, opened 2016-03-09
WIP Feedback

Maikon:
Here's some ideas for exposing these ideas in a simple way. Open to feedback of course.

### DIFF
--- a/lib/koans/05_functions.ex
+++ b/lib/koans/05_functions.ex
@@ -92,4 +92,13 @@ defmodule Functions do
   koan "You can 'capture' functions if you want to pass them around as values" do
     assert multiply_then_call(2, &square/1) == 100
   end
+
+  koan "functions can be combined elegantly with the pipe operator" do
+    result = "full-name"
+    |> String.split("-")
+    |> Enum.map(&(String.capitalize(&1)))
+    |> Enum.join(" ")
+
+    assert result == "Full Name"
+  end
 end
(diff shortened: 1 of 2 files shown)


lib/koans/05_functions.ex
```diff
@@ -92,4 +92,13 @@ defmodule Functions do
   koan "You can 'capture' functions if you want to pass them around as values" do
     assert multiply_then_call(2, &square/1) == 100
   end
+
+  koan "functions can be combined elegantly with the pipe operator" do
```
ukutaht: :+1:

lib/koans/09_pattern_matching.ex
```diff
@@ -5,6 +5,16 @@ defmodule PatternMatching do
     assert match?(1, 1)
   end
 
+  koan "a pattern can change" do
```
ukutaht: Don't really see where the blanks would go in these two. Unfortunately I don't have a better suggestion currently either...

felipesere:
Is there a way to expose how the pipe works? 
It's important to know that it puts the result of the previous function as the *first* argument of the next function.

lib/koans/09_pattern_matching.ex
felipesere: can you do 
```assert ^a == :__``` ?

Maikon: You can indeed do `assert ^a == :__`

Maikon:
I'm not sure how else to expose the pipe idea Felipe. I think the above (manipulating a string) it's the most basic thing I came up with. The other thing is people will roughly know that `map` takes two args and will figure out the same for the other functions having gone through most of the koans in this module before they get to the pipe. That's sort of part of the self-discovery process I guess :neutral_face:

PR #41, opened 2016-03-10
Test our koans

felipesere:
Sample how we could test our koans.

The principle is _simple-ish_:
create two functions: with no arguments used during regular koan evaluation and one with a single argument called `answer` for testing.

The function with the argument has a modified AST which replaces the `:__` with `argument`.

In the tests we then have a runner that calls that function with a proper argument, i.e. the answer to the koan.


felipesere:
@Maikon @ukutaht have a look before we proceed to test the remaining Modules.

Let's also think about what data structure we want for the answers. Right now its a simple list.

PR #44, opened 2016-03-14
Extract answers

felipesere:
* [x] Equalities (on master already)
* [x] Lists (on master already)
* [x] Maps
* [x] Strings
* [x] Functions
* [x] Enum
* [x] Arithmetic
* [x] Structs
* [x] Pattern Matching
* [x] Processes
* [x] Tasks
* [x] Tuples

### DIFF
--- a/lib/koans/12_tuples.ex
+++ b/lib/koans/12_tuples.ex
@@ -21,20 +21,12 @@ defmodule Tuples do
     assert Tuple.insert_at({:a, "hi"}, 1, :new_thing) == :__
   end
 
-  koan "...avoid falling of the edge" do
-    assert_raise :__, fn ->  Tuple.insert_at({:a, "hi"}, 12, :new_thing) end
-  end
-
   koan "add things at the end" do
     assert Tuple.append({"Huey", "Dewey"}, "Louie") == :__
   end
 
   koan "or also remove them" do
     assert Tuple.delete_at({:this, :is, :not, :awesome}, 2) == :__
-  end
-
-  koan "you can't delete what you don't have" do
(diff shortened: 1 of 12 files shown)


lib/koans/12_tuples.ex
```diff
@@ -0,0 +1,43 @@
...
+  koan "you can change individual elements of a tuple" do
+    assert put_elem({:a, "hi"}, 1, "bye") == :__
+  end
+
+  koan "you can also simply extend a tuple with new stuff" do
+    assert Tuple.insert_at({:a, "hi"}, 1, :new_thing) == :__
+  end
+
+  koan "...avoid falling of the edge" do
+    assert_raise :__, fn ->  Tuple.insert_at({:a, "hi"}, 12, :new_thing) end
```
ukutaht: Not sure about having a blank there.. How would the user know to fill this blank in? I guess if you have this open: http://elixir-lang.org/docs/stable/elixir/Tuple.html#insert_at/3. Are the koans supposed to be completed alongside the documentation?

I guess the overarching question is.. should I be able to figure out what the blank is by applying logic and reasoning or am I meant to swim in documentation as I'm doing these? I don't have an answer to it, just wanted to mention this.

felipesere: Tough call.  Should we put a hint on the name or even drop it altogether?
As I think through it, this kind of negative-case is close to the useless-edge.
You learn naturally about exceptions as they happen :smile:

ukutaht: :+1: Let's get rid of this then.

PR #50, opened 2016-03-20
Improve some descriptions a little bit

ukutaht:


lib/koans/05_lists.ex
```diff
@@ -9,10 +9,6 @@ defmodule Lists do
     assert List.last([1, 2, 3]) == :__
   end
 
-  koan "Diversity is embraced" do
```
felipesere: While the name is poor, I'd keep the the sentiment of the koan.
Maybe something along the lines

`koan "Lists can store anything throw at them" do`

ukutaht: Ah, I didn't realise that this is also the only koan with `++`. Will put it back in

felipesere:
LGTM.

PR #54, opened 2016-04-04
Introduce --koan=<name> to run a specific koan module.

felipesere:
The default is still to run `Equalities`.

If the name of the module is not known, an error and possible modules are shown in red:

screen shot 2016-04-04 at 20 58 59

@Maikon @ukutaht

### DIFF
--- a/lib/runner.ex
+++ b/lib/runner.ex
@@ -15,10 +15,11 @@ defmodule Runner do
   ]
 
   def run do
-    run(Equalities)
+    Options.initial_koan
+    |>run
   end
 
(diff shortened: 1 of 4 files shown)


lib/runner.ex
```diff
@@ -15,10 +15,11 @@ defmodule Runner do
   ]
 
   def run do
-    run(Equalities)
+    Options.initial_koan
+    |>run
   end
 
-  def run(start_module) do
+  def run(start_module) when start_module in @modules do
```
ukutaht: nice :ok_hand:

PR #63, opened 2016-04-09
Koan per process

felipesere:
This PR is necessary to build good Agents and Supervisior koans.
It changes the way we run koans from single process for all modules, to a individual process per _koan_.
That way, anything process that is started with a _link_ will be torn down after the koan.

It also splits _executing_ a koan from _running all koans_. The former has no dependencies, the latter depends on `Display` and `Options`.

A large chunk of credit goes to @marosluuce :+1: 

@ukutaht 
@Maikon

### DIFF
--- a/lib/execute.ex
+++ b/lib/execute.ex
@@ -1,36 +1,32 @@
 defmodule Execute do
   def run_module(module) do
-    run_until_failure(module, &run_koan/2)
-  end
-
-  def run_until_failure(module, callback) do
-    Enum.reduce_while(module.all_koans, :passed, fn(it, _acc) ->
-      result = callback.(module, it)
-      if result == :passed do
-        {:cont, :passed}
-      else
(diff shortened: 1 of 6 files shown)


lib/execute.ex
```diff
@@ -0,0 +1,48 @@
...
+    receive do
+      :ok   -> :passed
+      error -> {:failed, error, module, name}
+    end
+  end
+
+  defp exec(module, name, args, parent) do
+    result = apply(module, name, args)
+    send parent, expand(result)
+    Process.exit(self(), :kill)
```
marosluuce: I think you can drop this.

felipesere: Leaving it out fails the tests. My best guess is that it affects how children get cleaned up, if tall.

ukutaht:
whoa

PR #64, opened 2016-04-09
Attempt at making maps easier to go through

ukutaht:
@marosluuce In response to your feedback from Friday. If you think there's more to do here, feel free to push a commit on top of this :)

### DIFF
--- a/lib/koans/06_maps.ex
+++ b/lib/koans/06_maps.ex
@@ -1,65 +1,65 @@
 defmodule Maps do
   use Koans
 
+  @person %{
+    name: "Jon",
+    last_name: "Snow",
+    age: 27
+  }
+
   koan "Maps represent structured data, like a person" do
-    assert %{ :name => "Jon",
-              :last_name => "Snow",
-              :age => 27 } == %{ :name => :__,
-                                 :last_name => "Snow",
-                                 :age => 27 }
+    assert @person == %{name: :__,
+                        last_name: "Snow",
+                        age: 27 }
   end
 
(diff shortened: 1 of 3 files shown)


lib/koans/06_maps.ex
```diff
@@ -1,65 +1,65 @@
...
   end
 
   koan "You can get all the keys from the map" do
-    assert Map.keys(%{ :name => "Jon", :last_name => "Snow", :age => 27 }) |> Enum.sort == :__
+    assert Map.keys(@person) |> Enum.sort == :__
   end
 
   koan "Or you can also get all the values from it" do
-    assert Map.values(%{ :name => "Jon", :last_name => "Snow", :age => 27 }) |> Enum.sort == :__
+    assert Map.values(@person) |> Enum.sort == :__
```
felipesere: Throw this in with the `keys` koan

lib/koans/06_maps.ex
```diff
@@ -1,65 +1,65 @@
...
   end
 
   koan "Fetching a value returns a tuple with ok when it exists" do
-    assert Map.fetch(%{ :name => "Jon", :last_name => "Snow", :age => 27 }, :age) == :__
+    assert Map.fetch(@person, :age) == :__
   end
 
   koan "Or the atom :error when it doesnt" do
-    assert Map.fetch(%{ :name => "Jon", :last_name => "Snow", :age => 27 }, :family) == :__
+    assert Map.fetch(@person, :family) == :__
```
felipesere: Let's put the good case and the bad case next to each other.

PR #66, opened 2016-04-10
Improve enum koans

felipesere:
@ukutaht @Maikon @marosluuce


ukutaht:
We should be consistent about koans starting either with an uppercase letter or not. If I remember correctly we chose to start with uppercase, so there are a few here that need changing

PR #69, opened 2016-04-18
Remove arithmetic koans

ukutaht:
Removed arithmetic koans and added a few to equalities.

@Maikon @felipesere


felipesere:
Looks good. And CI is happy so :+1:

PR #73, opened 2016-04-23
Make sure blank is on the right and improve koans

felipesere:


lib/koans/12_agents.ex
```diff
@@ -27,10 +27,11 @@ defmodule Agents do
   end
 
   koan "Somebody has to switch off the light at the end of the day" do
-    Agent.start_link(fn() -> ["Milk"] end, name: __MODULE__)
+    {:ok, pid} = Agent.start_link(fn() -> ["Milk"] end, name: __MODULE__)
 
     result = Agent.stop(__MODULE__)
 
     assert result == ___
```
ukutaht: What if we got rid of this line and did

`:ok = Agent.stop(__MODULE__)`

up above?

PR #75, opened 2016-04-23
Some grammar changes

iamvery:
Couple small things. There was also "levelling the playing field" which my computer (and myself) thought was a typo. It seems like it also may be part of my North American english. Happy to send that along as well if it is indeed a fix.

### DIFF
--- a/lib/koans/07_pattern_matching.ex
+++ b/lib/koans/07_pattern_matching.ex
@@ -35,7 +35,7 @@ defmodule PatternMatching do
     assert third == ___
   end
 
-  koan "Strings come apart just a easily" do
+  koan "Strings come apart just as easily" do
     "Shopping list: " <> items = "Shopping list: eggs, milk"
 
     assert items == ___
--- a/lib/koans/10_processes.ex
+++ b/lib/koans/10_processes.ex
@@ -44,7 +44,7 @@ defmodule Processes do
     assert Process.alive?(pid) == ___
   end
 
-  koan "You can also terminate other processes than yourself" do
+  koan "You can also terminate processes other than yourself" do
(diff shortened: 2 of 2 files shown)


ukutaht:
Thank you for the contribution! 👍

iamvery:
You're welcome! Did you have any opinion/input on "levelling"? I'm curious 😄

ukutaht:
Hmm.. I wonder which spelling does Elixir itself favour? Most languages tend to prefer American English (`color` in CSS, `initialize` in Ruby) but I can't recall seeing anything in Elixir that reveals which spelling it prefers.

Overall I don't really mind which spelling we use as long as it's consistent. As most programming languages and programs themselves use American English I think we should just move to that.

So the short answer is that we should probably use 'leveling' and remove any other British spellings from the descriptions.

ukutaht:
For some background, we may have more British expressions/spellings in the koans because pretty much everyone who has contributed to this project lives in London.

iamvery:
It seems Elixir also favors American English (here's a reference to [color](http://elixir-lang.org/docs/stable/elixir/IO.ANSI.html#color/1)). I'm with you, I don't feel strongly either way :). Koans are usually a little whimsical which their wording, so it might even add a nice touch 😊. I'll let you all make those decisions! ❤️ 

I have found a couple other things to work on. I'll probably be sending a few PRs your way :soon:! I'm excited about the project 😃 https://twitter.com/iamvery/status/723874562446884864

ukutaht:
@iamvery Makes sense, I've pushed a change to use the American version ;)

There are still many areas to improve, PRs are greatly appreciated :)

iamvery:
Awesome! I just pushed one 💃

PR #77, opened 2016-04-24
Congratulate and exit when koans are complete

iamvery:
I worked through the entire set of koans and it took me a moment to realize that the last one wasn't stuck. It's because the process never exits (`:timer.sleep(:infinity)`). This restructures the code to print a nice little message when the entire run completes either initially or in the watcher.

I'd like to add tests around all these things, but the tests seem to be a little slim over the supporting code, so I opted to pass for now and test manually instead.

screen shot 2016-04-26 at 5 39 11 pm

### DIFF
--- a/lib/display.ex
+++ b/lib/display.ex
@@ -56,7 +56,6 @@ defmodule Display do
   def congratulate do
     format_green("\nYou have learned much. You must find your own path now.")
     |> IO.puts
-    exit(:normal)
   end
 
   def clear_screen do
--- a/lib/display.ex
+++ b/lib/display.ex
@@ -23,7 +23,7 @@ defmodule Display do
 
   def show_failure(failure, module, name) do
     IO.puts("Now meditate upon #{format_module(module)}")
-    IO.puts(progress_bar(Tracker.get))
+    IO.puts(progress_bar(Tracker.summarize))
     IO.puts(bar())
     IO.puts(name)
     IO.puts(format_failure(failure))
--- a/lib/display.ex
+++ b/lib/display.ex
@@ -53,6 +53,12 @@ defmodule Display do
     format_error(error) |> IO.puts
   end
 
+  def congratulate do
+    format_green("\nYou have learned much. You must find your own path now.")
+    |> IO.puts
+    exit(:normal)
+  end
+
   def clear_screen do
     if Options.clear_screen? do
       IO.puts(ANSI.clear)
@@ -95,6 +101,10 @@ defmodule Display do
 
   defp format_cyan(str) do
     Enum.join([ANSI.cyan, str, ANSI.reset], "")
+  end
+
+  defp format_green(str) do
+    Enum.join([ANSI.green, str, ANSI.reset], "")
   end
 
   defp format_module(module) do
--- a/lib/meditate.ex
+++ b/lib/meditate.ex
@@ -16,6 +16,8 @@ defmodule Mix.Tasks.Meditate do
     |> Tracker.start
     |> Runner.run
 
+    if Tracker.complete?, do: Display.congratulate
+
     receive do
       {:DOWN, _references, :process, watcher, _reason} -> nil
     end
--- a/lib/meditate.ex
+++ b/lib/meditate.ex
@@ -16,7 +16,10 @@ defmodule Mix.Tasks.Meditate do
     |> Tracker.start
     |> Runner.run
 
-    if Tracker.complete?, do: Display.congratulate
+    if Tracker.complete? do
+      Display.congratulate
+      exit(:normal)
+    end
 
     receive do
       {:DOWN, _references, :process, ^watcher, _reason} -> nil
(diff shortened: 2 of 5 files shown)


lib/runner.ex
```diff
@@ -30,14 +42,16 @@ defmodule Runner do
     modules
     |> Enum.drop_while( &(&1 != start_module))
     |> Enum.take_while( &(run_module(&1) == :passed))
+    |> done?
   end
 
+  defp done?(modules), do: modules == @modules
```
felipesere: `done?` Needs to be cleverer than that. If we start from somewhere in the middle of the koans (using `--koans=<Koan>`)  then `#drop_while(&(&1 != start_module))` will remove a bunch of modules.
That means `done?(module)` will never be `true`.

I am trying to come up with checks that will tell us when the user is done, but I am having a hard time...

iamvery: Dang you're absolutely right... What if we had a process that kept a running list of complete modules referred to that for doneness?

felipesere: That is exactly what I was working on this morning to display a progress bar :)
Build what you need in this branch and I will port my code if need be. 
If I do put together something in the next couple of hours, I'll open a PR for it.

iamvery: Sounds great!

iamvery: What's the desired behavior when you override the initial koan? In that case would it only run from that one to the last one?

iamvery:
@felipesere I reworked this. Hopefully it will be behave as needed now. I do realize that there is some overlap in functionality between my `Progress` module and your `Tracker` in #80. Since there is still some discussion going on over there, I opted to leave it so we could compare notes.

Something I felt repeatedly while putting this together is that it would be really great to get some unit tests up around the runner module. I suspect it has a little too much responsibility right now (which might make it awkward to test). I'll certainly give it a shot if I find myself with time to do so, but I didn't want it to hold up my pushing this and continuing to make progress.

iamvery:
@ukutaht @felipesere just pushed changes based on #80. Let me know what you think.

lib/meditate.ex
```diff
@@ -15,7 +16,11 @@ defmodule Mix.Tasks.Meditate do
     |> Tracker.start
     |> Runner.run
 
-    :timer.sleep(:infinity)
+    if Tracker.complete?, do: Display.congratulate
+
+    receive do
+      {:DOWN, _references, :process, watcher, _reason} -> nil
```
iamvery: I should probably pin (`^`) the `watcher` variable here, yes?

lib/tracker.ex
```diff
@@ -19,6 +19,11 @@ defmodule Tracker do
     end)
   end
 
+  def complete? do
+    {total, completed} = Agent.get(__MODULE__, &(&1))
+    total == Enum.count(completed)
```
iamvery: In my previous implementation, I actually kept a full list of the modules rather than just the count. For some reason, this worries me slightly that we might accidentally end up counting the wrong thing or something. Any thoughts on that?

iamvery:
Pretty neat to see how close the changes we made separately were to each other 😄

lib/tracker.ex
ukutaht: I guess this is a question for @felipesere. The progress seems to work pretty well so I trust the tracker at this point :)

lib/tracker.ex
```diff
@@ -19,6 +19,11 @@ defmodule Tracker do
     end)
   end
 
+  def complete? do
+    {total, completed} = Agent.get(__MODULE__, &(&1))
```
iamvery: There is a bit of duplication here and in `get/0`. I might suggest we refactor `get/0` to return the data itself and update `summarize/0` to be based on `get/0` rather than the other way around.

lib/display.ex
```diff
@@ -53,6 +53,12 @@ defmodule Display do
     format_error(error) |> IO.puts
   end
 
+  def congratulate do
+    format_green("\nYou have learned much. You must find your own path now.")
+    |> IO.puts
+    exit(:normal)
```
ukutaht: Would it be possible to let something else call `exit`? It doesn't feel right for the display to manage the process lifecycle but I may be missing something.

iamvery: Yeah, I thought so too, but it [wasn't unprecedented](https://github.com/elixirkoans/elixir-koans/blob/master/lib/display.ex#L10). Either way, I'll update this PR.

ukutaht:
This is looking good to me :+1:. What about you @felipesere ? Could you give it another look and merge if happy?

iamvery:
@ukutaht do you want me to do the `Tracker.get/0` refactor here, or in another PR?

ukutaht:
@iamvery Actually yes, that sounds like a good idea.

iamvery:
done!

lib/display.ex
```diff
@@ -53,6 +53,11 @@ defmodule Display do
     format_error(error) |> IO.puts
   end
 
+  def congratulate do
+    format_green("\nYou have learned much. You must find your own path now.")
```
felipesere: I would either pull the `IO.puts` onto the same line or pipe everywhere in the file, just for consistency.

felipesere:
Other than a tiny style thing with `IO.puts` this looks pretty cool 😎 .

@ukutaht @iamvery future notes:
* we have to remove `exit(:normal)` from `valid_koan?`
* we should push all the colour formatting into its own tiny little module

PR #78, opened 2016-04-24
Add trailing commas to multi-line lists and maps

iamvery:
This is _super_ pedantic, but I noticed an inconsistency in this style in the code, and I happen to prefer it this way 😆 (makes the git history cleaner when appending elements).

Feel free to ignore! 😅💦


felipesere:
Good point! Thank you! 👍

PR #79, opened 2016-04-24
Modify lists and maps style

fabientownsend:
This is a tiny modification!
I have updated the maps and lists to make the code look more consistent based on the documentation.

- [List documentation](http://elixir-lang.org/docs/stable/elixir/List.html)
- [Map documentation](http://elixir-lang.org/docs/stable/elixir/Map.html)


felipesere:
Cela me semble correct 👍

PR #80, opened 2016-04-24
Add a progress bar

felipesere:
*Updated*

Here is my new stab, with simpler code and better handling of `--koan=Agents`:

screen shot 2016-04-25 at 09 34 14

I wonder if we should a) add colors b) add a way to tell that you skipped X amount of koans.

### DIFF
--- a/lib/display.ex
+++ b/lib/display.ex
@@ -2,8 +2,7 @@ defmodule Display do
   alias IO.ANSI
   @current_dir File.cwd!
   @no_value :ex_unit_no_meaningful_value
-  @window_width 40
-  @half_window 20
+  @progress_bar_length 30
 
   def invalid_koan(koan, modules) do
     koans_names = module_names(modules)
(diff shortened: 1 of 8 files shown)


ukutaht:
What do you think about getting rid of the 'Considering' message? It was there to (sort of) indicate progress.

felipesere:
This still has an issue when you hit the half way mark 😦

lib/display.ex
```diff
@@ -22,11 +24,54 @@ defmodule Display do
...
+    progress_notification = "(#{current}/#{total})"
+
+    if current/total < 0.5 do
+      left_progress(progress_notification, progress)
+    else
+      right_progress(progress_notification, progress)
+    end
+  end
+
+  defp caluculate_progress(current, total) do
```
ukutaht: Typo here

lib/display.ex
```diff
@@ -2,6 +2,8 @@ defmodule Display do
   alias IO.ANSI
   @current_dir File.cwd!
   @no_value :ex_unit_no_meaningful_value
+  @window_width 40
```
ukutaht: It may be a bit misleading to call this @window_width. That name makes me think that it somehow dynamically calculates the width of the terminal window. How about just @progress_width or @progress_bar_width?

lib/display.ex
```diff
@@ -2,6 +2,8 @@ defmodule Display do
...
+  @half_window 20
```
ukutaht: I believe this can be defined as `@half_window @window_width / 2`

lib/tracker.ex
```diff
@@ -0,0 +1,24 @@
...
+    |> summirize
+  end
+
+  def completed(koan) do
+    Agent.update(__MODULE__, fn({total, completed}) ->
+      {total, MapSet.put(completed, koan)}
+    end)
+  end
+
+  def summirize({total, completed}) do
```
ukutaht: Typo

lib/execute.ex
```diff
@@ -1,12 +1,18 @@
 defmodule Execute do
-  def run_module(module) do
+  def run_module(module, callback \\ fn(result, koan) -> nil end) do
```
ukutaht: Is this default argument for tests? Would it look worse if we passed a dummy function in tests instead of making it the default on production side?

lib/meditate.ex
```diff
@@ -2,13 +2,31 @@ defmodule Mix.Tasks.Meditate do
   use Mix.Task
   alias Options
 
+  @modules [
```
ukutaht: Is this used?

test/tracker_test.exs
```diff
@@ -0,0 +1,25 @@
...
+
+  test "can be notified of completed koans" do
+    Tracker.start(@sample_modules)
+    Tracker.completed(:"Hi there")
+    assert Tracker.get == %{total: 2, current: 1}
+  end
+
+  test "multiple comletions of the same koan count only once" do
+    Tracker.start(@sample_modules)
+    Tracker.completed(:"Hi there")
```
ukutaht: 2 Should be enough to show that it doesn't count double :)

ukutaht:
Just saw a problem with this: when I misspell a koan module like this
```bash
$ mix meditate --clear-screen --koan=Agenst
```

It doesn't give me an error message as it used to, it just kinda hangs which makes it feel broken

ukutaht:
Otherwise looks good :+1: lovely!

@iamvery Once this is merged, are you happy to update the end congratulations to work with this code? Your PR and this one are conflicting pretty heavily so we have to pick one to merge first :)

test/tracker_test.exs
felipesere: Done.

felipesere:
@ukutaht I'll look into the missing error tonight. That must have happened when I moved the responsibilities of the runner around...

iamvery:
@ukutaht absolutely ❤️

felipesere:
@ukutaht and @iamvery I fixed the error when passing the wrong koan name would not fail early.
Have another look and merge when happy :)

ukutaht:
❤️ 💛 💚 💜 💙

PR #82, opened 2016-04-25
Struct koans now calls '|' operator by correct name: 'cons'

Zanadar:
Little change to remove some ambiguity in the koans. `|>` is he pipe operator, but `|`, in `erlang/elixir` is the 'cons' operator (short for constructor).


felipesere:
Good catch! I guess my brain was in Unix mode 😄

PR #85, opened 2016-04-26
Attempt clarifying the pin operator

ukutaht:
@RabeaGleissner @felipesere Do you think this is a better demonstration of the pin operator?

### DIFF
--- a/lib/koans/07_pattern_matching.ex
+++ b/lib/koans/07_pattern_matching.ex
@@ -8,12 +8,14 @@ defmodule PatternMatching do
   koan "A pattern can change" do
     a = 1
     a = 2
-    assert (a == 2) == ___
+    assert a == ___
   end
 
   koan "A pattern can also be strict" do
     a = 1
-    assert ^a = ___
+    assert_raise ___, fn() ->
+      ^a = 2
+    end
   end
 
   koan "Patterns can be used to pull things apart" do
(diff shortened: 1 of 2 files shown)


lib/koans/07_pattern_matching.ex
```diff
@@ -8,12 +8,14 @@ defmodule PatternMatching do
...
     a = 1
     a = 2
-    assert (a == 2) == ___
+    assert a == ___
   end
 
   koan "A pattern can also be strict" do
     a = 1
-    assert ^a = ___
+    assert_raise ___, fn() ->
```
felipesere: I think figuring out that a `MatchError` was raised is a little too hard.
How about we add `assert raise_happend(fn() -> ... end)`?
`raise_happend` would just return true if something raised, false otherwise...

lib/koans/07_pattern_matching.ex
```diff
@@ -8,12 +8,14 @@ defmodule PatternMatching do
...
     a = 1
     a = 2
-    assert (a == 2) == ___
+    assert a == ___
   end
 
   koan "A pattern can also be strict" do
     a = 1
-    assert ^a = ___
+    assert_raise MatchError, fn() ->
```
ukutaht: @felipesere How about this?

RabeaGleissner: To me this makes sense! @ukutaht

felipesere: Much better! I like it!

PR #86, opened 2016-04-26
Map.fetch! -> Map.fetch in 06_structs.ex

Troush:
I think is better to continue use safe ```Map.fetch``` until we familiarize users with dangerous function concept.
Not every body comes from ruby background, with knowledge of ```?``` and ```!``` function notations.


ukutaht:
@Troush Agreed, thanks for spotting that :)

The build is failing because we have tests with answers for all koans. This change forces us to change the provided answer in `test/koans/structs_koans_test.exs`.

Troush:
Update :)

Fixed ```structs_koans_test.exs```

ukutaht:
Awesome, thanks!

PR #87, opened 2016-04-26
Error is raised if an answer isn't provided for each koan

iamvery:
Currently, answers are zipped with the list of koans which might allow examples to go untested. This admittedly inelegant solution ensures that you at least provide an answer for each koan. It should provide a good sanity check while TDDing koans.

Honestly I'm not a huge fan of this solution, but it was simple enough to implement and serves to start a conversation about answers. 😃


ukutaht:
Makes sense 👍

PR #88, opened 2016-04-26
Add test for agents koans

iamvery:
Noticed that the agents koans didn't have tests. Here they are 👍


felipesere:
Oh darn, I must have swallowed when I split our huge test file into multiple small ones. 
Good catch!

PR #89, opened 2016-04-27
Assert koan test results

iamvery:
Currently failing koans cause the parent process to hang while waiting for results. This sends failures back to the parent process and assert that everything passed.

### DIFF
--- a/lib/koans.ex
+++ b/lib/koans.ex
@@ -33,8 +33,12 @@ defmodule Koans do
     single_var = Blanks.replace(body, Macro.var(:answer, Koans))
     quote do
       def unquote(name)(answer) do
-        unquote(single_var)
-        :ok
+        try do
+          unquote(single_var)
+          :ok
+        rescue
+          e in _ -> e
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/koans.ex
```diff
@@ -44,8 +48,12 @@ defmodule Koans do
...
       def unquote(name)({:multiple, answers}) do
         converted = List.to_tuple(answers)
-        unquote(multi_var)
-        :ok
+        try do
+          unquote(multi_var)
+          :ok
+        rescue
+          e in _ -> e
+        end
```
iamvery: Some duplication here...

test/test_helper.exs
```diff
@@ -19,6 +22,10 @@ defmodule TestHarness do
     end
   end
 
+  defp check_results(results) do
+    Enum.each results, &(assert &1 == :passed)
```
felipesere: I kinda want the parens around `.each(...)` 😸

iamvery: Done! Curious, what's your style on this? I tend to like omitting parens on multiline calls that have anonymous functions.

```elixir
Enum.each list, fn item ->
  # something
  # else
end
```

For single-line, I don't have any clear preference. In this case, I think I opted to leave them out to avoid the double parens at the end.

What's your metric?

felipesere: I can see the appeal... for some reason I've always been cautious with the parens.
@ukutaht What's your take?

ukutaht: Parens. I like to visually disambiguate function calls from variables.

PR #92, opened 2016-05-02
Update various koans

iamvery:
I made a bunch of notes working through the set of koans early last week. I'm going to push changes here for discussion. I'm happy to spin off separate PRs if necessary 😅

### DIFF
--- a/lib/koans/07_pattern_matching.ex
+++ b/lib/koans/07_pattern_matching.ex
@@ -69,7 +69,7 @@ defmodule PatternMatching do
   def make_noise(%{type: "dog"}), do: "Woof"
   def make_noise(_anything), do: "Eh?"
 
-  koan "Functions can declare what kind of arguments they accept" do
+  koan "Functions perform pattern matching on their arguments" do
     cat = %{type: "cat"}
     dog = %{type: "dog"}
     snake = %{type: "snake"}
--- a/lib/koans/07_pattern_matching.ex
+++ b/lib/koans/07_pattern_matching.ex
@@ -110,4 +110,9 @@ defmodule PatternMatching do
     %Animal{name: name} = %Animal{kind: "dog", name: "Max"}
     assert name == ___
   end
+
+  koan "Structs will even match with a regular map" do
+    %{name: name} = %Animal{kind: "dog", name: "Max"}
+    assert name == ___
+  end
 end
--- a/lib/koans/07_pattern_matching.ex
+++ b/lib/koans/07_pattern_matching.ex
@@ -5,7 +5,12 @@ defmodule PatternMatching do
     assert match?(1, ___)
   end
 
-  koan "A pattern can change" do
+  koan "A value can be bound to a variable" do
+    a = 1
+    assert a == ___
+  end
+
+  koan "A variable can be rebound" do
     a = 1
     a = 2
     assert a == ___
--- a/lib/koans/07_pattern_matching.ex
+++ b/lib/koans/07_pattern_matching.ex
@@ -49,7 +49,7 @@ defmodule PatternMatching do
     assert items == ___
   end
 
-  koan "Patterns show what you really care about" do
+  koan "Maps support partial pattern matching" do
     %{make: make} = %{type: "car", year: 2016, make: "Honda", color: "black"}
 
     assert make == ___
--- a/lib/koans/07_pattern_matching.ex
+++ b/lib/koans/07_pattern_matching.ex
@@ -55,6 +55,12 @@ defmodule PatternMatching do
     assert make == ___
   end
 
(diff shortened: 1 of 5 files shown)


lib/koans/07_pattern_matching.ex
```diff
@@ -5,13 +5,18 @@ defmodule PatternMatching do
     assert match?(1, ___)
   end
 
-  koan "A pattern can change" do
+  koan "A value can be bound to a variable" do
+    a = 1
+    assert a == ___
+  end
+
+  koan "A variable can be rebound" do
```
iamvery: I felt like it was more accurate to use the language "bound" in this case. However, it seems it needed it's own introduction before talking about "rebound".

lib/koans/07_pattern_matching.ex
```diff
@@ -5,13 +5,18 @@ defmodule PatternMatching do
...
     a = 1
     a = 2
     assert a == ___
   end
 
-  koan "A pattern can also be strict" do
+  koan "A variable can be pinned to be prevent it from being rebound" do
```
iamvery: It seemed that this description unnecessarily danced around the terminology "pin".

lib/koans/07_pattern_matching.ex
```diff
@@ -44,12 +49,18 @@ defmodule PatternMatching do
...
   end
 
-  koan "Patterns show what you really care about" do
+  koan "Maps support partial pattern matching" do
     %{make: make} = %{type: "car", year: 2016, make: "Honda", color: "black"}
 
     assert make == ___
   end
 
+  koan "Lists must match exactly" do
```
iamvery: It seemed nice to draw the contrast between how matching works with maps vs. lists.

lib/koans/07_pattern_matching.ex
```diff
@@ -44,12 +49,18 @@ defmodule PatternMatching do
...
+    assert_raise ___, fn ->
```
iamvery: At this point, the reader has already experienced a `MatchError`, so it seems fair to ask for it here.

lib/koans/07_pattern_matching.ex
```diff
@@ -99,4 +110,9 @@ defmodule PatternMatching do
     %Animal{name: name} = %Animal{kind: "dog", name: "Max"}
     assert name == ___
   end
+
+  koan "Structs will even match with a regular map" do
+    %{name: name} = %Animal{kind: "dog", name: "Max"}
+    assert name == ___
```
iamvery: Trying to get them thinking "huh, structs really are a lot like maps" which sets us up to illustrate exactly how they're _not_ later.

lib/koans/08_functions.ex
```diff
@@ -50,7 +50,7 @@ defmodule Functions do
   def the_length(0), do: "It was zero"
   def the_length(number), do: "The length was #{number}"
 
-  koan "For those individual one-offs, you can even guard on the arguments themselves" do
+  koan "For simpler cases, pattern matching is effective" do
```
iamvery: I felt that this was more accurately pattern matching than guarding.

ukutaht:
Nice work :+1: What parts are still WIP?

iamvery:
Well I suppose nothing that's here is still WIP, but I do have more notes to go through. I'll remove WIP if you want to go ahead and merge what you see. Can open new PRs for the rest. 😃

ukutaht:
I see. Yeah, another PR for the rest would be nice. The smaller the PRs the better :)

I'm happy with this, @felipesere why don't you give it a look and merge if all good.

lib/koans/10_processes.ex
```diff
@@ -2,17 +2,17 @@ defmodule Processes do
   use Koans
 
   koan "Tests run in a process" do
-    assert Process.alive?(self()) == ___
+    assert Process.alive?(self) == ___
```
iamvery: I noticed in a screencast I was watching today that Chris McCord tends to include empty parens on function calls with no arguments. This goes a little against my aesthetic, but I wanted to mention it because perhaps my statement about this "not being idiomatic" isn't entirely true. What are you thoughts?

ukutaht: I think the jury is still out... I tend to use parens as well but not all the time. For `self` I probably wouldn't use them but I have no good reason for why :)

iamvery: I tend to go with Avdi's [barewords](http://devblog.avdi.org/2012/10/01/barewords/) semantic for this kind of thing. Maybe not as applicable to Elixir code as it is to Ruby, but it's my history 😄

iamvery:
Rebased to fix merge conflict.

Issue #94, opened 2016-05-03
assert_receive not compatible with strategy for multiple answers


iamvery:
This isn't a very big deal, but I wanted to report it in case it became relevant.

# Problem

The current strategy for answering a koan with multiple blanks during test is not compatible with `assert_receive`.

As we've found in #93, the expression given to `assert_receive` is used as a match in a `receive` call that is generated by the macro. The current strategy for inserting multiple answers is to replace blanks (`___`) with generated expressions, e.g.  `elem(answers, 0)`. However the expression `elem(list, index)` is not a valid match, which breaks compilation.

```
» mix test
Compiled lib/runner.ex

== Compilation error on file lib/koans/10_processes.ex ==
** (CompileError) lib/koans/10_processes.ex:18: illegal pattern
    (stdlib) lists.erl:1337: :lists.foreach/2
    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6
```

Again, this really isn't an issue until we have koans missing multiple answers _and_ using `assert_receive` (or any other macro that expects its expression to be used in a match), but it seems worth reporting while the problem is understood. Also as you can see the error is cryptic and hard to debug.

PR #95, opened 2016-05-03
Restructure/simplify process koans

iamvery:
While working through the process koans, I found some opportunity to improve the lessons themselves. The changeset is a little noisy, but the story should be relatively clear if read commit-wise.

1. Reword some descriptions
1. Change initial tests to establish baseline: everything's a process, create processes with `spawn`, receive messages with `receive`
1. Get rid of `wait/0` function since it's only used in one place (and might be mistaken for something built-in)
1. Simplify some of the messaging examples by eliminating extra process layers

Also worth mentioning that these examples still suffer from #93, but that should be fixed (or fixable) once we come to a conclusion there.

What do you think?

### DIFF
--- a/lib/koans/10_processes.ex
+++ b/lib/koans/10_processes.ex
@@ -69,7 +69,11 @@ defmodule Processes do
                         {:EXIT, _pid, reason} -> send parent, {:exited, reason}
                       end
     end)
-    wait()
+
+    receive do
+      :ready -> true
+    end
+
     Process.exit(pid, :random_reason)
 
     assert_receive ___
@@ -117,11 +121,5 @@ defmodule Processes do
      end)
 
     assert_receive ___
-  end
-
-  def wait do
-    receive do
-      :ready -> true
-    end
   end
 end
--- a/lib/koans/10_processes.ex
+++ b/lib/koans/10_processes.ex
@@ -80,16 +80,16 @@ defmodule Processes do
   end
 
   koan "Trying to quit normally has no effect" do
-    pid = spawn(fn -> receive do
-                      end
-                end)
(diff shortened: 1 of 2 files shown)


ukutaht:
You know I've been thinking that some of these koans, especially:
* koan "Trying to quit normally has no effect"
* koan "Exiting normally yourself on the other hand DOES terminate you"
* koan "Parent processes can be informed about exiting children, if they trap and link"
* koan "If you monitor your children, you'll be automatically informed for their depature"

They seem overly detailed about signal trapping and process monitoring without really teaching what to use processes for.

On the other hand I really like what you've done [here](https://github.com/iamvery/elixir-koans/blob/master/about_processes.exs), especially with the echo loop (and the one before that). What are your thoughts on porting some of those koans over? I feel like they're much better at communicating the idea behind processes without getting into excruciating detail.

ukutaht:
Otherwise, I'm all for all the changes and improved descriptions :+1:

iamvery:
Thanks for the feedback! I'm all for porting stuff over from iamvery/elixir-koans. It's actually on my list of things todo, i.e. run through those and think about what might be useful here. Would you want me to try to do that as part of this PR or another one after merge?

ukutaht:
Awesome. I'll merge this and we can work on porting stuff over in the next PR.

iamvery:
I just realized that porting over https://github.com/iamvery/elixir-koans/blob/7705d743acf3720f75852eb6badfe57f428a3481/about_processes.exs#L50-L64 is going to be problematic cause #94. We could try to address that or just revert to regular `receive` calls for that example.

PR #98, opened 2016-05-04
Rearrange, rm, and add process examples

iamvery:
Following #95, this does a little more house cleaning on the processes koan. Additionally, a couple new examples are added.

- receiving multiple messages
- using a process to store state


lib/koans/10_processes.ex
```diff
@@ -26,15 +45,71 @@ defmodule Processes do
...
+    send pid, {self, "o"}
+    send pid, {self, "hai"}
+
+    receive do
+      msg -> assert msg == ___
+    end
+
+    receive do
+      msg -> assert msg == ___
+    end
```
iamvery: Can't use `assert_receive` here because of #94

lib/koans/10_processes.ex
```diff
@@ -26,15 +45,71 @@ defmodule Processes do
...
+    send pid, {self, :get}
+    receive do
+      value -> assert value == ___
+    end
+
+    send pid, {self, :set, "bar"}
+    send pid, {self, :get}
+    receive do
+      value -> assert value == ___
+    end
```
iamvery: Can't use `assert_receive` here either, because #94

lib/koans/10_processes.ex
```diff
@@ -26,15 +45,71 @@ defmodule Processes do
...
+  end
+
+  koan "Use tail recursion (calling a function as the very last statement) to receive multiple messages" do
+    pid = spawn &yelling_echo_loop/0
+
+    send pid, {self, "o"}
+    send pid, {self, "hai"}
+
+    receive do
```
ukutaht: Could we interleave the calls to `send` and `receive` here? So instead of `send-send-receive-receive` we could have `send-receive-send-receive`. Feels more intuitive to me, not sure if you agree.

lib/koans/10_processes.ex
```diff
@@ -11,13 +11,32 @@ defmodule Processes do
...
-  koan "You can send messages to processes" do
+  koan "You can kill processes other than yourself" do
+    pid = spawn(fn -> receive do end end)
+
+    assert Process.alive?(pid) == ___
+    Process.exit(pid, :kill)
+    assert Process.alive?(pid) == ___
+  end
+
+  koan "However, trying to exit normally has no effect" do
```
ukutaht: What do you think about removing this?

lib/koans/10_processes.ex
```diff
@@ -26,15 +45,71 @@ defmodule Processes do
...
 
+  def yelling_echo_loop do
+    receive do
+      {caller, value} ->
+        send caller, String.upcase(value)
+        yelling_echo_loop
+    end
+  end
+
+  koan "Use tail recursion (calling a function as the very last statement) to receive multiple messages" do
```
ukutaht: Do we need to explain tail recursion here? How about just naming this 'Use recursion to receive multiple messages' or something like that?

ukutaht:
I merged the `assert_receive` fix, can update these to use it now.

lib/koans/10_processes.ex
```diff
@@ -26,15 +45,71 @@ defmodule Processes do
...
+
+    receive do
+      msg -> assert msg == ___
+    end
+
+    receive do
+      msg -> assert msg == ___
+    end
+
+    Process.exit(pid, :kill)
```
ukutaht: Would the koan leak this process if we didn't kill it here?

lib/koans/10_processes.ex
iamvery: I don't have any strong opinion. My thoughts when looking back at these were general "eh, not so awful I suppose". Whatever you'd like 👍

lib/koans/10_processes.ex
iamvery: I'm sort of on the fence about it. On one hand, the example certainly works without the explanation. On the other hand, it's extremely important to this style of code.

That said, it doesn't _actually_ explain tail recursion, it just throws the term out there. Perhaps a better way to revealing this would be to have this example (without mention of tail recursion) and then another example that illustrates why tail recursion is important... Not sure what that would look like.

Thoughts?

lib/koans/10_processes.ex
iamvery: I liked this arrangement only because it really illustrates the mailbox-like (or queue, FIFO if you will) behavior of Erlang messaging. Maybe that's better explained in it's own example?

lib/koans/10_processes.ex
iamvery: I _think_ so, or at least it did in my koans project. Maybe it's not needed here, because of https://github.com/elixirkoans/elixir-koans/blob/master/lib/execute.ex#L36 (which I actually have a note about, because it wasn't clear to me why it was needed at first)

ukutaht: Not sure about that line but I'm also afraid of leaking processes. I know that it used to be a problem but I'm not sure if it still is. 

In theory since all koans run in their own process, I'd expect them to tear down everything they created once the koan exits. I don't know if this is how it currently works though because I don't fully comprehend the linking mechanism and which processes erlang decides to kill/keep alive.

I'd propose to keep this line in but investigate the overall issue. End game is to get to a point where koans can start child processes that get automatically cleaned up when the koan process exits. Shouldn't affect this PR though.

lib/koans/10_processes.ex
ukutaht: I can't really think of a way to teach about tail recursion using koans. The only visible effect of it is faster runtime and bounded memory usage – not really something we can make nice assertions about.

I see the point of at least mentioning tail recursion, though. It is an important concept and something that every functional programmer should understand. Maybe just throwing the term out there is enough. If users want to dig deeper they can go off on their own.

Could we at least get rid of the text in parentheses? To me the real lesson of the koan is that by recursing you can keep responding to messages. I'm afraid that the text in parentheses may distract users to think that the koan is _about_ tail recursion itself rather than the `receive -> recurse` idiom.

lib/koans/10_processes.ex
ukutaht: Let's remove it then. I don't see a lot of value in it.

lib/koans/10_processes.ex
ukutaht: I see, that makes sense but it wasn't really obvious at first sight. It's almost like the koan is trying to teach too much at the same time. I like your suggestion of having a different koan to demonstrate the message queue.

ukutaht:
Overall, some really awesome stuff here :heart: I feel like the process koans were lacking the most so far, glad to see them improving!

ukutaht:
Also, #94 was closed today so if you wanted to use multiple answers with `assert_receive` it should be working now 🎉

lib/koans/10_processes.ex
felipesere: If we want to show the mailbox nature of the messages, we could write our own version of `assert_receive` that takes an entire list (_aka mailbox).
Something like
```
send self(), :foo
send self(), :bar
assert_mailbox [:foo, :bar]
```

Is that an option worth considering?

iamvery: I don't like writing a custom `assert_mailbox`, because it might lead the reader to think that is an Elixir thing.

I'll write up a separate koan to illustrate the mailbox-like behavior.

lib/koans/10_processes.ex
felipesere: As far I can remember, only `spawn_link`ed stuff get destroyed when the parent dies.
Linked processes are bi-directional. Either one dies or is killed and the buddy dies too.

If we "just" `spawn` then we have to clean up the pids ourselves. If we use `spwan_link` then the fact that the koan-process get's killed also kills the processes within the koan. (sounds convoluted, I know)

lib/koans/10_processes.ex
iamvery: Sounds good :)

lib/koans/10_processes.ex
```diff
@@ -26,15 +37,62 @@ defmodule Processes do
...
+  def yelling_echo_loop do
+    receive do
+      {caller, value} ->
+        send caller, String.upcase(value)
+        yelling_echo_loop
+    end
+  end
+
+  koan "Use tail recursion to receive multiple messages" do
+    pid = spawn &yelling_echo_loop/0
```
ukutaht: We're using parens around `spawn` everywhere else, probably worth adding them here for consistency.

lib/koans/10_processes.ex
iamvery: I just confirmed this. So if we use `spawn_link` instead, we won't need the kill. Do it?

lib/koans/10_processes.ex
```diff
@@ -11,30 +11,87 @@ defmodule Processes do
...
+  def yelling_echo_loop do
+    receive do
+      {caller, value} ->
+        send caller, String.upcase(value)
+        yelling_echo_loop
+    end
+  end
+
+  koan "Use tail recursion to receive multiple messages" do
+    pid = spawn_link(&yelling_echo_loop/0)
```
iamvery: I went ahead and replace the `Process.exit` with `spawn_link`. Can revert if needed @felipesere @ukutaht

felipesere: Looks good.

ukutaht:
Good stuff :+1: Thanks for the quick updates

PR #100, opened 2016-05-04
Replace pipeline with composed function in maps koan

iamvery:
At this point, the learner hasn't seen functions, the Enum module, or
the pipe operator, so it feels like a little much to add that here. The
reason it's needed is that the order of keys/1 and values/1 is reverse
of how it's defined. This is due to an implementation detail in Erlang's
maps. We /could/ use this as a learning opportunity rather than sorting
them, but I'm not sure what could be said about that...

Other options include
- get rid of the example
- limit example to single key and value to avoid order problems
- use it as a learning opportunity (what do we learn?)

### DIFF
--- a/lib/koans/05_maps.ex
+++ b/lib/koans/05_maps.ex
@@ -38,7 +38,7 @@ defmodule Maps do
 
   koan "Can remove pairs by key" do
     without_age = Map.delete(@person, :age)
-    assert Enum.sort(Map.keys(without_age)) == ___
+    assert Map.has_key?(without_age, :age) == ___
   end
 
   koan "Can merge maps" do
--- a/lib/koans/05_maps.ex
+++ b/lib/koans/05_maps.ex
@@ -13,11 +13,6 @@ defmodule Maps do
                         age: 27 }
   end
 
-  koan "A map has keys and values" do
-    assert Enum.sort(Map.keys(@person)) == ___
-    assert Enum.sort(Map.values(@person)) == ___
-  end
-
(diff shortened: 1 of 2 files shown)


lib/koans/05_maps.ex
```diff
@@ -43,7 +43,7 @@ defmodule Maps do
 
   koan "Can remove pairs by key" do
     without_age = Map.delete(@person, :age)
-    assert Map.keys(without_age) |> Enum.sort == ___
+    assert Enum.sort(Map.keys(without_age)) == ___
```
ukutaht: Maybe we could replace this with [Map.has_key?/2](http://elixir-lang.org/docs/stable/elixir/Map.html#has_key?/2) so that we don't rely on the key order? I'm thinking something like:

```elixir
assert Map.has_key?(without_age, :age) == false
```

lib/koans/05_maps.ex
```diff
@@ -14,8 +14,8 @@ defmodule Maps do
   end
 
   koan "A map has keys and values" do
```
ukutaht: I would be okay with deleting this koan. It feels very clunky in the current form.

iamvery: Seems fine. It's only point is to introduce folks to some of the `Map` interface. Not like we're going to cover it all I suppose.

lib/koans/05_maps.ex
iamvery: Yeah, that's perfect 💫

iamvery:
updated

ukutaht:
👍

PR #101, opened 2016-05-04
Allow multiple answers in pinned context

### LINKED ISSUE #94: assert_receive not compatible with strategy for multiple answers

ukutaht:
Instead of accessing multiple answers via `elem(converted, id)` I've changed the koan macro to just destruct the answers in the test function signature. This way all answers become variables on their own and can be pinned in our special `assert_receive` handling. This PR doesn't include an example of using this but I've confirmed that it works with the `yelling_echo_loop` from #98.

closes #94

### DIFF
--- a/lib/koans.ex
+++ b/lib/koans.ex
@@ -43,11 +43,11 @@ defmodule Koans do
     end
   end
   defmacro generate_test_method(name, number_of_args, body) do
-    answer_placeholders = create_vars(number_of_args)
-    multi_var = Blanks.replace(body, answer_placeholders)
+    answer_vars = for id <- 1..number_of_args, do: Macro.var(String.to_atom("answer#{id}"), Koans)
+    multi_var = Blanks.replace(body, answer_vars)
+
     quote do
-      def unquote(name)({:multiple, answers}) do
-        converted = List.to_tuple(answers)
+      def unquote(name)({:multiple, unquote(answer_vars)}) do
         try do
           unquote(multi_var)
           :ok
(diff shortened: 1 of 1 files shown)


iamvery:
Nice! Great work :)

lib/koans.ex
```diff
@@ -43,11 +43,11 @@ defmodule Koans do
     end
   end
   defmacro generate_test_method(name, number_of_args, body) do
-    answer_placeholders = create_vars(number_of_args)
-    multi_var = Blanks.replace(body, answer_placeholders)
+    answer_vars = for id <- 1..number_of_args, do: Macro.var(String.to_atom("answer#{id}"), Koans)
```
iamvery: Seems like you could have left the function that generates these values rather than doing it inline. That said, I don't feel strongly that it really improves the readability (it's all really complicated regardless).

Issue #96, opened 2016-05-03
Autoreload doesn't work on macbook


marioosh:
I'm trying to make koans, but it shows me an error every time I save the edited file:

    ** (Code.LoadError) could not load /sandbox/elixir_koans/elixir-koans/lib/koans/05_maps.ex___jb_tmp___
    (elixir) lib/code.ex:662: Code.find_file/2
    (elixir) lib/code.ex:319: Code.load_file/2

I have Macbook Pro with Retina 2015, fresh installed, the newest Elixir(1.2.4) from Homebrew

marioosh:
I pulled into new folder and now it works, weird :/

marioosh:
Reopened: This issue is present when file is edited in IntelliJ IDEA with Elixir plugin.

ukutaht:
Thanks for reporting. It seems like IntelliJ is writing some temp files into the watched directory which the runner doesn't expect. We're looking into this

felipesere:
@marioosh can you pull the branch with `git fetch origin pull/102/head:intellij-fix && git checkout intellij-fix` and retry?

marioosh:
Works great !!!

2016-05-04 22:56 GMT+02:00 Felipe Seré <notifications@github.com>:

> @marioosh <https://github.com/marioosh> can you pull the branch with git
> fetch origin pull/102/head:intellij-fix && git checkout intellij-fix and
> retry?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/elixirkoans/elixir-koans/issues/96#issuecomment-217000288>
>



-- 
Mariusz Nosiński
507-183-661
mnosinski@gmail.com

PR #102, opened 2016-05-04
Strip off IntelliJ specific extension from reloaded files.

felipesere:
Should solve #96 by stripping away the IntelliJ specific suffix.

### DIFF
--- a/lib/watcher.ex
+++ b/lib/watcher.ex
@@ -3,20 +3,30 @@ defmodule Watcher do
 
   def callback(file, events)  do
     if Enum.member?(events, :modified) do
-      try do
-        Code.load_file(file)
-          |> Enum.map(&(elem(&1, 0)))
-          |> Enum.find(&Runner.koan?/1)
-          |> Runner.modules_to_run
-          |> Runner.run
-      rescue
-        e -> Display.show_compile_error(e)
-      end
+      reload(file)
 
       if Tracker.complete? do
         Display.congratulate
         exit(:normal)
       end
     end
+  end
+
+  defp reload(file) do
+    try do
+      file
+      |> normalize
+      |> Code.load_file
+      |> Enum.map(&(elem(&1, 0)))
+      |> Enum.find(&Runner.koan?/1)
+      |> Runner.modules_to_run
+      |> Runner.run
+    rescue
+      e -> Display.show_compile_error(e)
+    end
+  end
+
+  defp normalize(file) do
+    String.replace_suffix(file, "___jb_tmp___", "")
   end
 end


lib/watcher.ex
```diff
@@ -3,20 +3,30 @@ defmodule Watcher do
...
+      |> Enum.find(&Runner.koan?/1)
+      |> Runner.modules_to_run
+      |> Runner.run
+    rescue
+      e -> Display.show_compile_error(e)
+    end
+  end
+
+  defp normalize(file) do
+    String.replace_suffix(file, "___jb_tmp___", "")
```
iamvery: I feel like this would cause the file to be mistakenly rerun whenever an IntelliJ file is modified. Would it be desirable to just skip changes on these files? Or better yet, maybe only watch for changes to `.ex` files in that directory?

felipesere: Weirdly enough, no further events are triggered.
Initially I silently dropped files with the `__jb_tmp___` suffix but nothing happened...

ukutaht: The second option is a no-go because `exfswatch` requires a whole directory to be watched at once, the library doesn't provide a way to watch only specific extensions.

Maybe @felipesere can comment on the first suggestion.

iamvery: I played with this a bit. I think _new_ files aren't triggering the watcher, because it's only looking for `:modified` events. So running the task and `touch`ing new files wouldn't trigger anything. Is that approximately what you were doing?

iamvery: Could we just add a small guard that does something like `String.ends_with?(file, ".ex")`?

felipesere: I installed the Elixir IntelliJ plugin and was editing the koans with it and it worked like a charm.
The `mix meditate` itself was running in iTerm2.

Touching files should not be an issue there is no reason to do so in the koans.

iamvery: Right. I definitely agree that this solution works. My concern is that it's too specific to IntelliJ. What if some other user has another editor that drops temporary files into the directory? So I was thinking to come at it from the other direction. Perhaps only attempt to load Elixir files in that directory. Would it help if I pushed up the implementation I have in mind?

felipesere:
I tried just ran your version and for some reason I get *exactly* one modified event and nothing after that.

I also have this version that extracts a basename using a regex:
```
  defp normalize(file) do
    case Regex.run(~r/(.+ex).*/, file) do
      [_entire | [base | _reminder]] -> base
      _ -> raise "could not extract proper filename from "<>file
    end
  end
```

To be honest, I don't mind special casing a bunch of IDEs as long as it stays readable. I don't think that many more IDEs create funky files. And if they do, all we need is something like
```
def normalize(file) do
  file
  |> String.replace_suffix(file, "___jb_tmp___", "")
  |>String.replace_suffix(file, "other_ending...", "")
end
```
If that does get out of control, we can think of something more generic.

So I vote for the current code in this PR :)

iamvery:
I didn't see the "one event" behavior, but there's no telling. I threw it together pretty quickly 😆 .

I'm fine with this. Worse case scenario is we have to update a couple more times for other's environments, and we learn something :blush:.

PR #103, opened 2016-05-04
Add meditate shortdoc to display task in mix list

slavik7:
This change makes meditate task visible when user runs `mix help`.

### DIFF
--- a/lib/meditate.ex
+++ b/lib/meditate.ex
@@ -2,6 +2,8 @@ defmodule Mix.Tasks.Meditate do
   use Mix.Task
   alias Options
 
+  @shortdoc "Start the koans"
+
   def run(args) do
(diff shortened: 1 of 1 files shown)


lib/meditate.ex
```diff
@@ -2,6 +2,8 @@ defmodule Mix.Tasks.Meditate do
   use Mix.Task
   alias Options
 
+  @shortdoc "Start Elixir Koans runner"
```
iamvery: Maybe "runner" is an implementation detail here. Perhaps just "Start the koans"

slavik7: Right point. Updated it.

ukutaht:
Awesome, thank you! :+1:

PR #105, opened 2016-05-05
Atoms koan

iamvery:
Porting over atoms koans from iamvery/elixir-koans. Unfortunately there are a couple that don't work with the way answers are implemented here. Worth trying to address? Throw them away?


lib/koans/03_atoms.ex
```diff
@@ -0,0 +1,55 @@
...
+    assert String.to_atom("atomized") == ___
+    assert Atom.to_string(:stringified) == ___
+  end
+
+  koan "Atoms are often used as keys, because they're faster than strings" do
+    map = %{name: "Jay"}
+    list = [name: "Jay"]
+
+    assert map[:name] == ___
+    assert list[:name] == ___
```
iamvery: This lesson doesn't feel nearly as useful without the next two.

lib/koans/03_atoms.ex
```diff
@@ -0,0 +1,55 @@
...
+    assert :false == ___
+  end
+
+  koan "Modules are also atoms" do
+    assert is_atom(String) == ___
+    assert :"Elixir.String" == ___
+    assert :"Elixir.String".upcase("hello") == ___
+  end
+
+  koan "Atoms are used to access Erlang" do
```
ukutaht: Can we remove this in favor of having a whole koan file for erlang interop? Ref #19

lib/koans/03_atoms.ex
```diff
@@ -0,0 +1,55 @@
...
+  koan "It is surprising to find out that booleans are atoms" do
+    assert is_atom(true) == ___
+    assert is_atom(false) == ___
+    assert :true == ___
+    assert :false == ___
+  end
+
+  koan "Modules are also atoms" do
+    assert is_atom(String) == ___
+    assert :"Elixir.String" == ___
```
ukutaht: I suspect the intention here is to fill it out like this: `assert :"Elixir.String" == String`. I doubt that people will actually do that, though, because it's not obvious behaviour in Elixir and it's much easier to make it pass with `assert :"Elixir.String" == :"Elixir.String"`.

Thoughts?

lib/koans/03_atoms.ex
```diff
@@ -0,0 +1,55 @@
...
+    adam = :human
+    assert adam == ___
+  end
+
+  koan "Strings can be converted to atoms, and vice versa" do
+    assert String.to_atom("atomized") == ___
+    assert Atom.to_string(:stringified) == ___
+  end
+
+  koan "Atoms are often used as keys, because they're faster than strings" do
```
ukutaht: I wonder if this fits better in the Maps or Keyword koans? The users haven't seen maps or keyword lists at this point so it may be confusing...

ukutaht:
Allowing expressions in answers seems pretty hard to do :) I'd happily merge this PR without the problematic koans.

lib/koans/03_atoms.ex
iamvery: Sure!

lib/koans/03_atoms.ex
iamvery: Ya, totally. Maybe we could actually use the atom instead of just referencing it?

```elixir
koan "Modules are also atoms" do
  assert is_atom(String) == ___
end

koan "You can even access functions from the atom directly, but this is not typical" do
  yelling = :"Elixir.String".upcase("hi")
  assert yelling == "HI"
end
```

iamvery:
We were able to allow expressions in answers (see https://github.com/iamvery/elixir-koans/pull/21/files#r60759404), but doing so on this project would require major changes to have answers are inserted in test. Probably not worth it for these examples, but could be a fun endeavor regardless.

lib/koans/03_atoms.ex
iamvery: That seems legit.

iamvery: Since we can't illustrate the dot access (or lack there of) yet, maybe just remove it. The point "are faster than" isn't really illustrated anyway.

iamvery:
Updated. Rebased and addressed the things we talked about.

lib/koans/03_atoms.ex
ukutaht: I like this much better 👌

Issue #21, opened 2016-01-15
Koans about macros


ukutaht:

iamvery:
Consider including examples of `do...else...end` being a keyword list. See https://github.com/elixirkoans/elixir-koans/pull/106#discussion_r62267413

PR #106, opened 2016-05-05
Keyword list koan

iamvery:
Porting over from iamvery/elixir-koans (https://github.com/iamvery/elixir-koans/blob/master/about_lists_and_maps.exs)


iamvery:
rebased

lib/koans/06_keyword_lists.ex
```diff
@@ -0,0 +1,62 @@
...
+    transform = fn str, opts ->
+      if opts[:upcase], do: String.upcase(str)
+    end
+
+    assert transform.("good", upcase: true) == ___
+  end
+
+  def foo(kw_list), do: kw_list
+
+  koan "Actually function bodies are a sneaky use of keyword lists" do
```
ukutaht: I had a really hard time figuring out what's going on here. I understand that it is indeed an interesting implementation detail but do beginners need to know this to get going with the language? I'm afraid that it ends up just confusing people who are just picking up the language.

iamvery: I'd buy that even though I do think it's _super_ cool. Maybe we could move it to the macro koan (#21) since becomes applicable when dealing with ASTs?

ukutaht: Agreed, let's make it part of macro/AST koans.

iamvery:
updated (removed)

ukutaht:
:+1: I'm ready to merge this, just needs a rebase.

iamvery:
rebased

felipesere:
LGTM 👍

PR #108, opened 2016-05-05
Swap order of zipped lists to create learning opportunity

iamvery:
Zipping atoms and numbers creates an opportunity for the learner to
realize "💡 I just made a keyword list!"

### DIFF
--- a/lib/koans/10_enums.ex
+++ b/lib/koans/10_enums.ex
@@ -55,9 +55,9 @@ defmodule Enums do
   end
 
   koan "Zip-up in pairs!" do
-    numbers = [1, 2, 3]
     letters = [:a, :b, :c]
-    assert Enum.zip(numbers, letters) == ___
+    numbers = [1 ,2 ,3]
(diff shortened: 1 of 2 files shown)


iamvery:
rebased

ukutaht:
:+1:
Needs another rebase, though :)

iamvery:
rebased

iamvery:
Honestly, adding koans sort of sucks with the naming scheme we've got for these files. We might want to consider doing away with the numbering or come up with something else 😖

iamvery:
With all the conflicts, it was tough to be certain I wasn't throwing away a previous change

ukutaht:
The reason I named them with numbers initially was because files are ordered alphabetically in text-editors, IDEs, `ls`, you name it. I though it makes it very clear which file to go to next. 

I'm open to suggestions of course :)

iamvery:
Yeah, it totally makes sense! Which reminds me... I've forgotten to add the modules to the runner on all these new koans 😆

PR #114, opened 2016-05-07
Improvements suggestions

riacataquian:
Here are various improvements regarding styleguides and wordings. Would love feedbacks!

### DIFF
--- a/lib/koans/06_maps.ex
+++ b/lib/koans/06_maps.ex
@@ -2,13 +2,13 @@ defmodule Maps do
   use Koans
 
   @person %{
-    name: "Jon",
+    first_name: "Jon",
     last_name: "Snow",
     age: 27,
   }
 
   koan "Maps represent structured data, like a person" do
-    assert @person == %{name: ___,
+    assert @person == %{first_name: ___,
                         last_name: "Snow",
                         age: 27 }
   end
@@ -42,7 +42,7 @@ defmodule Maps do
   end
 
   koan "Can merge maps" do
-    assert Map.merge(%{name: "Jon"}, %{last_name: "Snow"}) == ___
+    assert Map.merge(%{first_name: "Jon"}, %{last_name: "Snow"}) == ___
   end
 
   koan "When merging, the last map wins" do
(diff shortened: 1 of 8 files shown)


lib/koans/09_functions.ex
```diff
@@ -47,20 +47,20 @@ defmodule Functions do
     assert bigger(4, 27) == ___
   end
 
-  def the_length(0), do: "It was zero"
-  def the_length(number), do: "The length was #{number}"
+  def get_number(0), do: "The number was zero"
+  def get_number(number), do: "The number was #{number}"
```
felipesere: I don't know what I was thinking when I named these functions. 👍

felipesere:
Awesome! Thanks for your suggestions.
Do you mind looking through the other koans to check for inconsistent indentation and comma-in-list-missplacements?

lib/koans/11_processes.ex
```diff
@@ -81,8 +81,8 @@ defmodule Processes do
   koan "Processes can be used to hold state" do
     initial_state = "foo"
     pid = spawn(fn ->
-      state(initial_state)
-    end)
+            state(initial_state)
```
ukutaht: I have to say that to my eyes it's more readable without the change :shrug:

lib/koans/06_maps.ex
```diff
@@ -2,13 +2,13 @@ defmodule Maps do
   use Koans
 
   @person %{
-    name: "Jon",
+    first_name: "Jon",
```
ukutaht: :+1:

riacataquian:
@felipesere Sure, no problem.

lib/koans/11_processes.ex
riacataquian: Is that so? I'm willing to undo it. Do you want me to? :)

ukutaht: Yeah, I'd say let's undo it if you're OK with that :)

riacataquian:
Also, I have a question @felipesere and @ukutaht. Should we be using Elixir's convention when referring to functions?

Please see: https://github.com/elixirkoans/elixir-koans/blob/master/lib/koans/13_agents.ex#L14
I'd suggest it to be: `To update the state of an agent, use the Agent.update/2 function`. I'm having some doubts. Thanks!

felipesere:
I am not overly picky on `Agent.update/2` vs. `update`. 😸 
Once we have more infrastructure in place, I'd prefer to push the descriptions (of the modules) and the individual koans more towards story-telling than API-describing. 📚

riacataquian:
Sure @felipesere thanks for the input! 👍

PR #115, opened 2016-05-10
Adds an @intro attribute to display when running a module

felipesere:
To help build more of a story into the koans,
I've added a module attribute called `@intro`.

That intro will be displayed the very first time a module is accessed.

screen shot 2016-05-10 at 23 05 11

### DIFF
--- a/lib/display.ex
+++ b/lib/display.ex
@@ -21,8 +21,18 @@ defmodule Display do
   defp name(module), do: name(Atom.to_string(module))
 
   def show_failure(failure, module, name) do
-    intro(module, Tracker.visited)
     IO.puts(format(failure, module, name))
+  end
+
+  def format(failure, module, name) do
+    """
+    #{intro(module, Tracker.visited)}
+    Now meditate upon #{format_module(module)}
+    #{progress_bar(Tracker.summarize)}
+    ----------------------------------------
+    #{name}
+    #{format_failure(failure)}
+    """
   end
 
   defp intro(module, modules) do
@@ -36,17 +46,6 @@ defmodule Display do
     message <> "\n"
     |> Colours.green
     |> IO.puts
-  end
-
-
-  def format(failure, module, name) do
-    """
-    Now meditate upon #{format_module(module)}
-    #{progress_bar(Tracker.summarize)}
-    ----------------------------------------
-    #{name}
-    #{format_failure(failure)}
-    """
   end
 
   def progress_bar(%{current: current, total: total}) do
--- a/lib/display.ex
+++ b/lib/display.ex
@@ -20,6 +20,18 @@ defmodule Display do
   defp name("Elixir." <> module), do: module
   defp name(module), do: name(Atom.to_string(module))
 
+  def intro(module) do
+    show_intro(module.intro)
+    module
+  end
+
+  defp show_intro(""), do: nil
+  defp show_intro(message) do
+    message <> "\n"
+    |> Colours.green
+    |> IO.puts
+  end
+
   def show_failure(failure, module, name) do
     IO.puts(format(failure, module, name))
   end
--- a/lib/display.ex
+++ b/lib/display.ex
@@ -20,9 +20,15 @@ defmodule Display do
   defp name("Elixir." <> module), do: module
   defp name(module), do: name(Atom.to_string(module))
 
-  def intro(module) do
-    show_intro(module.intro)
-    module
+  def show_failure(failure, module, name) do
+    intro(module, Tracker.visited)
+    IO.puts(format(failure, module, name))
+  end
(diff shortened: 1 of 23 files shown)


felipesere:
@ukutaht @Maikon @gemcfadyen @iamvery would love your feedback :)

lib/display.ex
```diff
@@ -21,9 +21,24 @@ defmodule Display do
   defp name(module), do: name(Atom.to_string(module))
 
   def show_failure(failure, module, name) do
+    intro(module, Tracker.visited)
     IO.puts(format(failure, module, name))
   end
 
+  defp intro(module, modules) do
+    if ! module in modules do
```
ukutaht: Could you clarify the order of operations here? It's possible to read it as `if !(module in modules)` and `if (!module) in modules`. I don't know if this is possible, but I'd imagine you can also do `if module not in modules` or something like that.

lib/koans/01_equalities.ex
```diff
@@ -1,6 +1,13 @@
 defmodule Equalities do
   use Koans
 
+  @intro  """
```
ukutaht: I like the message here but the last sentence seems off grammatically. To me it seems like it should read 'The path laid in front of you is but one of many.' Changes are:
`layed` -> `laid` (http://public.wsu.edu/~brians/errors/layed.html)
`infront` -> `in front`
`us` -> `is`

ukutaht:
Huge ❤️ overall.

lib/koans/01_equalities.ex
felipesere: Fixed, though I would probably suggest we improve 'the story' in different PR. 
That line was just an example 😸

lib/koans.ex
```diff
@@ -82,11 +82,32 @@ defmodule Koans do
...
     end
   end
+
+  defp koans(env) do
+    env.module
+    |> Module.get_attribute(:koans)
+    |> Enum.reverse
+  end
+
+  defp extract_into(env) do
```
iamvery: Did you mean to name this function `extract_intro`?

lib/koans.ex
```diff
@@ -82,11 +82,32 @@ defmodule Koans do
...
-    koans = Module.get_attribute(env.module, :koans) |> Enum.reverse
+    koans = koans(env)
+    intro = extract_into(env)
     quote do
       def all_koans do
         unquote(koans)
       end
+
+      def intro do
+        unquote(intro)
```
iamvery: Is all this metaprogramming needed? Couldn't his function just be something like

```elixir
def intro, do: @intro
```

felipesere: If every module has `@intro` then you are right.
I could flick this around and make sure that `use Koans` adds the attribute with "" if there is nothing defined in the module itself.

What do you think?

iamvery: It seems like the culmination of this would be that all koans have intros and the metaprogramming is no longer needed, so I like the idea of including empty @intro to avoid the metaprogramming.

felipesere: I'll just add a blank `@intro ""` to all the other koans and kill the metaprogramming :)

felipesere:
@iamvery Added @intro to all modules and removed the metaprogramming.

lib/display.ex
```diff
@@ -21,9 +21,24 @@ defmodule Display do
   defp name(module), do: name(Atom.to_string(module))
 
   def show_failure(failure, module, name) do
+    intro(module, Tracker.visited)
```
iamvery: Why is the intro printed on failure? I'm sure it works as you expect, but something about that seems weird.

felipesere: We only ever print failures. Success "roll on through" to the next koan.
I was torn between showing the intro in failure and adding a clause around display itself.

What if I rename it to `maybe_intro`? To sneaky? 😄

iamvery: If it works as it is maybe don't even worry about it. I suppose you could add a note or a comment or something that suggests a refactor in the future if you wanted to.  It's probably not a big deal

felipesere:
@iamver you were absolutely spot on. I moved the function call into `format_failure`. The nice thing is that `format_failure` now shows exactly what is going to be printed - on a high level.

I also got the feeling that I want to split `Display` into multiple modules. Some that prep `Strings` and `Display` that prints it. What do you think?

If you are happy with this PR, I'll merge it 🙇

iamvery:
@felipesere nice! This reads a lot better. Intention revealing 😃. As far as splitting the modules, I don't feel deeply opinionated. If it makes it more clear what does what, it makes sense. On the other hand, if these functions are only in support of the `Display` then they can probably live on as private functions there.

PR #116, opened 2016-05-12
Adds koans for protocols

gemcfadyen:
@felipesere @ukutaht @Maikon 

Hi all, I've added a koan for protocols. Please take a look and see what you think.

### DIFF
--- a/lib/koans/15_protocols.ex
+++ b/lib/koans/15_protocols.ex
@@ -1,0 +1,39 @@
+defmodule Protocols do
+  use Koans
+
+  defprotocol School do
+    def enrol(name)
+  end
+
+  defimpl School, for: SecondarySchool do
+    def enrol(name) do
+      "#{name} enrolled at secondary school"
+    end
+  end
+
+  defimpl School, for: DanceSchool do
+    def enrol(name) do
+      "#{name} enrolled for dance"
+    end
+  end
+
+  defmodule EveningClass do end
+
+  @schools %{
+    :secondary => School.SecondarySchool,
+    :dance => School.DanceSchool,
+    :evening => EveningClass
+  }
+
+  koan "Commonality is the secret" do
+    school = Map.get(@schools, :dance)
+    assert school.enrol("Emily") == ___
+  end
+
+  koan "When there is a difference there is no way" do
+    assert_raise ___, fn ->
+      school = Map.get(@schools, :evening)
+      school.enrol("bob")
+    end
+  end
+end
--- a/test/koans/protocols_koans_test.exs
+++ b/test/koans/protocols_koans_test.exs
@@ -1,0 +1,13 @@
+defmodule ProtocolsTests do
+  use ExUnit.Case
+  import TestHarness
+
+  test "Protocols" do
+    answers = [
+      "Emily enrolled for dance",
+      UndefinedFunctionError
+    ]
+
+    test_all(Protocols, answers)
+  end
+end
--- a/lib/runner.ex
+++ b/lib/runner.ex
@@ -14,6 +14,7 @@ defmodule Runner do
     Processes,
     Tasks,
     Agents,
+    Protocols,
   ]
 
   def koan?(koan), do: Enum.member?(@modules, koan)


lib/koans/15_protocols.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Protocols do
+  use Koans
+
+  defprotocol School do
+    def enrol(name)
```
iamvery: Please pardon me if this is an international spelling, but I believe the correct spelling is "enroll" (with two Ls).

lib/koans/15_protocols.ex
```diff
@@ -0,0 +1,39 @@
...
+  defmodule EveningClass do end
+
+  @schools %{
+    :secondary => School.SecondarySchool,
+    :dance => School.DanceSchool,
+    :evening => EveningClass
+  }
+
+  koan "Commonality is the secret" do
+    school = Map.get(@schools, :dance)
```
iamvery: Using a map in this case actually feels like an unnecessary indirection. Any problem with just doing

```
School.DanceSchool.enroll("Emily")
```

lib/koans/15_protocols.ex
```diff
@@ -0,0 +1,39 @@
...
+  defimpl School, for: DanceSchool do
+    def enrol(name) do
+      "#{name} enrolled for dance"
+    end
+  end
+
+  defmodule EveningClass do end
+
+  @schools %{
+    :secondary => School.SecondarySchool,
```
iamvery: It appears the `SecondarySchool` is never used.

lib/koans/15_protocols.ex
```diff
@@ -0,0 +1,39 @@
...
+    :evening => EveningClass
+  }
+
+  koan "Commonality is the secret" do
+    school = Map.get(@schools, :dance)
+    assert school.enrol("Emily") == ___
+  end
+
+  koan "When there is a difference there is no way" do
+    assert_raise ___, fn ->
```
iamvery: I don't think it's fair to assume the learner will know what error belongs here. I'm also not sure the description of this example explains the bit of learning very well. What is the "difference"?

test/koans/protocols_koans_test.exs
```diff
@@ -0,0 +1,13 @@
+defmodule ProtocolsTests do
+  use ExUnit.Case
+  import TestHarness
+
+  test "Protocols" do
+    answers = [
+      "Emily enrolled for dance",
+      UndefinedFunctionError
```
iamvery: Ah interesting, I would have expected a `Protocol.UndefinedError`

iamvery:
Thanks for taking a crack at this! I'm really excited about expanding the coverage of this project so it helps more people :). The examples left a little to be desired from me. I like the examples used in the [official docs](http://elixir-lang.org/getting-started/protocols.html). It might not be wrong to model this koan after that. It might also be really cool to illustrate use of `@derive`, `Any` and `@fallback_to_any` though they are admittedly advanced topics. Finally for illustrating the "protocol not implemented" case, I feel like we need to come up with a more enlightening exercise. I don't have a specific suggestion at the moment unfortunately...

Great work! Thank you ❤️

lib/koans/15_protocols.ex
gemcfadyen: [Double 'l'  is the American spelling](http://grammarist.com/spelling/enrol-enroll/) and else where one 'l' is usedI am happy to change to use the American spelling if preferred.

lib/koans/15_protocols.ex
gemcfadyen: You can absolutely use what you suggested.  I used the map to show how Protocols could be used polymorphically as I felt that is where you would get benefit from using them in a project.

lib/koans/15_protocols.ex
gemcfadyen: Again you are right, however I was trying to make the user think a little bit. There are two 'Schools' that implement the Protocol, and I was hoping the user might have to figure out the answer. Directly calling the module as you suggested earlier or just having one entry in the map, makes the answer quite obvious and doesn't really show the polymorphic advantage. 

I could add another test similar to the one for DanceSchool, for SecondarySchool? Then both would be used..

lib/koans/15_protocols.ex
gemcfadyen: I was also unsure whether to have this koan. It was to demonstrate what happens if you try to call a function in a place where you expect polymorphism to take place and the module does not implement the interface.

If there is benefit in showing what happens when something doesn't adhere to the expected Protocol we can keep the koan, if not then I am happy to remove this koan.

gemcfadyen:
I've had a look at the Protocol examples on the documentation you pointed too. Seems a shame to just copy paste those examples directly. I'll play around with the School example next week using Structs as in the input and see if they are better. 

Thanks for the feedback!

lib/koans/15_protocols.ex
felipesere: Hahahah. You'll notice this throughout the Koans. @gemcfadyen @RabeaGleissner @ukutaht @Maikon  and I are all based in London.

test/koans/protocols_koans_test.exs
felipesere: Indeed interesting.

lib/koans/15_protocols.ex
iamvery: Haha, I figured that was the case. Listen ya'll are lucky to get something legible out of this Alabama boy (see what I did there? :wink:)

@ukutaht and I had a conversation about this before: https://github.com/elixirkoans/elixir-koans/pull/75. I'm not deeply opinionated one way or the other 😺

lib/koans/15_protocols.ex
iamvery: I agree, it totally makes sense to have two modules implement a protocol to show the full advantage of them. I do think we need to come up with another example that involves the other... 🤔

lib/koans/15_protocols.ex
iamvery: The more I think about this, I'm not sure the map is actually illustrating polymorphism... I mean the modules could just as easy implement a common interface and be used interchangeably, right? That's probably why we're not getting the expected "protocol error" below. I'll have to think a little more about this. I haven't done much with protocols yet.

lib/koans/15_protocols.ex
iamvery: There is _totally_ a benefit! We just need to nail down the approach.

test/koans/protocols_koans_test.exs
iamvery: https://github.com/elixirkoans/elixir-koans/pull/116/files/d93b7fa0f53c01a4426eb4f6dcbf03f68131536c#r63178434

iamvery:
I do agree that it seems shameful to just take the example outright. I suppose I just appreciate the practical nature of it. Maybe we can come up with something a little less contrived?

Thank you for receiving the feedback! I can come on a little strong, so I'm glad it didn't upset you. Your contributions are greatly appreciated! And your work is great ❤️

lib/koans/15_protocols.ex
iamvery: Ah hah! I realized what the problem is. You don't utilize a protocol by calling functions on the module, you do it by calling the function on the protocol.

So instead of saying `School.DanceSchool.enrol("Jay")`, you would say `School.enrol(student)` where student would be a value with a protocol defined for it. Does that make sense?

Take a closer look at http://elixir-lang.org/getting-started/protocols.html. The protocol is exercised by calling Blank.blank?/1 with different types of arguments.

```
iex> Blank.blank?(0)
false
iex> Blank.blank?([])
true
iex> Blank.blank?([1, 2, 3])
false
iex> Blank.blank?("hello")
** (Protocol.UndefinedError) protocol Blank not implemented for "hello"
iex(3)> Blank.String.blank?("no")
** (UndefinedFunctionError) undefined function Blank.String.blank?/1 (module Blank.String is not available)
    Blank.String.blank?("no")
```

Calling methods directly on the specific type, e.g. Blank.Atom.blank?/1, defeats the purpose.

iamvery: And that makes sense, right? Polymorphism is processing things differently based on their type. If you define multiple functions with the same interface that can be used interchangeably, that's more along the lines of duck typing.

iamvery:
Just saw this in #elixir-lang

> protocols are useful if you dont control all of the code, for example if you have json library, you cant change the json library when you want to add new types to encode

Might help as a more concrete example

gemcfadyen:
Closing as updated functionality in: https://github.com/elixirkoans/elixir-koans/pull/124

PR #117, opened 2016-05-13
Extract smaller modules from display.

felipesere:
To improve testabilty, I have extracted four new modules out of the Display.
The key thing is that the smaller modules don't do any IO, they only generate strings.
All the `IO.puts` are collected in the `Display` class.

I also put a little indirection around colours to disable them during testing.

@ukutaht @iamvery @Maikon @gemcfadyen


lib/display/intro.ex
```diff
@@ -0,0 +1,16 @@
...
+    if not module in modules do
+      show_intro(module.intro)
+    else
+      ""
+    end
+  end
+
+  def show_intro(message) do
+    message <> "\n"
+    |> Colours.green
```
ukutaht: Should this be using the `Paint` module as well?

PR #118, opened 2016-05-16
Make clearing the screen the default

felipesere:
As multiple people have said, we should make clearing the screen the default.

@ukutaht @iamvery @Maikon @gemcfadyen what is your opinion?


gemcfadyen:
Think clearing the screen by default is a good idea. It will allow the user to focus on the question/error at hand.

Maikon:
@felipesere I agree. Like @gemcfadyen said, it keeps the focus on the next immediate error only.

PR #120, opened 2016-05-18
Move pinning to the back of the kaon and enhance example

felipesere:
I thought we needed a more explicit example about pattern matching and pinning.

@ukutaht @Maikon @iamvery @gemcfadyen @RabeaGleissner

### DIFF
--- a/lib/koans/10_pattern_matching.ex
+++ b/lib/koans/10_pattern_matching.ex
@@ -111,7 +111,7 @@ defmodule PatternMatching do
     assert a == ___
   end
 
-  koan "A variable can be pinned to prevent it from being rebound" do
+  koan "A variable can be pinned to use its value when matching instead of binding to a new value" do
     pinned_variable = 1
 
     example = fn
@@ -124,14 +124,20 @@ defmodule PatternMatching do
     assert example.(3) == ___
   end
 
-  koan "Pinning works on anywhere one would match, including 'case'" do
+  koan "Pinning works anywhere one would match, including 'case'" do
     pinned_variable = 1
-    other_variable = 1
-    result = case other_variable do
+    result = case 1 do
                ^pinned_variable -> "same"
                other -> "different #{other}"
              end
 
     assert result == ___
+  end
+
+  koan "Trying to rebind a pinned variable will result in an error" do
+    a = 1
+    assert_raise MatchError, fn() ->
+      ^a = ___
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/koans/10_pattern_matching.ex
```diff
@@ -117,4 +99,39 @@ defmodule PatternMatching do
...
+      (^pinned_variable) -> "The number One"
+      (2) -> "The number Two"
+      (number) -> "The number #{number}"
+    end
+    assert example.(1) == ___
+    assert example.(2) == ___
+    assert example.(3) == ___
+  end
+
+  koan "Pinning works on anywhere one would match, including 'case'" do
```
iamvery: Either remove the word "on" or reword to "work on anything"

lib/koans/10_pattern_matching.ex
```diff
@@ -117,4 +99,39 @@ defmodule PatternMatching do
...
+    pinned_variable = 1
+    other_variable = 1
+    result = case other_variable do
```
iamvery: I sort of feel like the `other_variable` is distracting here.

lib/koans/10_pattern_matching.ex
```diff
@@ -7,24 +7,6 @@ defmodule PatternMatching do
...
-
-  koan "A variable can be rebound" do
-    a = 1
-    a = 2
-    assert a == ___
-  end
-
-  koan "A variable can be pinned to prevent it from being rebounded" do
-    a = 1
-    assert_raise MatchError, fn() ->
```
iamvery: I do still sort of like this to illustrate what happens when a pinned match fails. Maybe introduce this before some of the more advanced examples?

lib/koans/10_pattern_matching.ex
felipesere: One of these days I'll actually check my spelling before pushing...

lib/koans/10_pattern_matching.ex
felipesere: Should I just do 'case 1 do'?

iamvery: Ya, but that's a bit meh, huh?

iamvery: Ya, but that's a bit meh, huh?

lib/koans/10_pattern_matching.ex
felipesere: Done, but at the bottom of the koan. The idea is to introduce how to use it (well) and then what happens when it fails.

iamvery: Sounds great! 🎉

PR #122, opened 2016-05-24
Add koans for integers

msurekci:
Hi, I just thought I would add some koans for the Integer module. It is a pretty slim module so I didn't add many and tried to keep it simple. 

I am thinking about adding further koans for using the optional base parameter. That would be for using 2 for Binary and 16 for Hexidecimal.


felipesere:
Thank you so much for your PR!
Do you think there are more 'numerical' operations that might be worth while? Something from the `Float` module maybe?
My thinking is that I'd prefer to cover all of the interesting functions on numbers in one module rather than splitting them. 

Also, what do you think about the order? I feel like `Integers` should be introduced fairly early...

msurekci:
Thanks for the feedback. I would be glad to include more functionality from the `Float ` module. There is also a `Range` module do you think it can be included in this one module?

I also, agree that it should be introduced fairly early. I just threw it out here to get some feedback. Do you think it would be good to have it after the strings koans?

felipesere:
Yeah,  go ahead and also add the `Range` module.

@ukutaht @Maikon any comments?

lib/koans/04_numbers.ex
```diff
@@ -0,0 +1,93 @@
+defmodule Numbers do
+  require Integer
+  use Koans
+
+  @intro "Numbers"
```
felipesere: I really like the titles for your koans and I know we haven't put tooo much effort into the `@intro ....` text. It would be awesome if this koans were a shining example :) Do you mind adding a little text, funny and explanatory about numbers in Elixir? 2-3 lines should suffice.

lib/koans/04_numbers.ex
```diff
@@ -0,0 +1,93 @@
...
+    assert 4 in range == ___
+    assert 10 in range == ___
+    assert 0 in range == ___
+  end
+
+  koan "Is this a range?" do    
+    assert Range.range?(1..10) == ___
+    assert Range.range?(0) == ___
+  end
+
```
felipesere: A little superfluous empty lines has sneaked in...

felipesere:
Awesome! Thank you!

msurekci:
Thanks for your support Felipe. This was my first open source contribution :)

iamvery:
Great work @msurekci! What an awesome way to make your mark on OSS 🎉

PR #123, opened 2016-05-26
Adding new koans for map sets

msurekci:
Hi, I noticed there was an issue to create koans for MapSets. I took it upon myself to add some. Appreciate any feedback.

Also, I didn't change the order yet. I thought I would get some input from you folks.


lib/koans/17_map_sets.ex
```diff
@@ -0,0 +1,62 @@
+defmodule MapSets do
+  use Koans
+
+  @intro "My name is Set, MapSet"
+
+  @set MapSet.new([1, 2, 3, 4, 5])
+
+  koan "Does this value exist in the map set?" do
```
iamvery: I sort of feel like the first introduction to MapSet should be something like "It's a collection, like a list", and then the second lesson be "But it's different in that each element is distinct, like a set". What we have here gets off to a slow start and sort of makes me want to ask "wait what is this thing for?"

Does that make sense? Maybe I'm crazy? 😆

felipesere:
Thanks for the PR and picking up the MapSet.

I would add a koan about the features that make a MapSet interesting.
For example, add the same thing three times, but show that the size doesn't change.

lib/koans/17_map_sets.ex
msurekci: Yup, makes sense. I will make some changes :)

felipesere:
@msurekci  You'll need to rebase this with latest `master` due to other PRs.

felipesere:
LGTM! @iamvery @Maikon @gemcfadyen any final comments?

lib/koans/10_map_sets.ex
```diff
@@ -0,0 +1,62 @@
+defmodule MapSets do
+  use Koans
+
+  @intro "My name is Set, MapSet."
+
+  @set MapSet.new([1, 2, 3, 4, 5])
+
+  koan "I am merely another collection but, you can perform some operations on me" do
+    new_set = MapSet.new(@set, fn x -> 3 * x end)
```
iamvery: I like the description of this, but the first example still feels a little loaded to me. It's illustrating

1. how to initialize a set
1. that you can provide a function to change the set
1. that you can check a set for a member

I tend to think of the learning progression as
1. a set is like a list - show a set initialized with list and how it's like a list
1. however a set does not allow duplicates - show a set initialized with dups and show that they aren't in the set
1. further lessons about sets such as init with operation

That said, I don't feel so strongly that I'd recommend against merging 😄 . We can always improve later!

felipesere: Now that I think about it, I think you are right @iamvery.

how about we add a very first koan that shows that 'its kinda like list'?
Create a MapSet and then use any fun `Enum` function on it like you would with a list.
And then show duplication and then proceed to any of the fun MapSet specific APIs?

@msurekci thanks for sticking with us ❤️

msurekci: @felipesere I like the idea. The Enum koans come in at number 13 (on master). I placed this before the Enums koans. Would it be OK to use the Enum function before it is introduced or maybe this koan should come after the Enum koans?

felipesere: I think we might want to move both Enum and Functions up the list at some point, so I wouldn't be too fuzzed about it.

gemcfadyen:
LGTM

msurekci:
@felipesere @iamvery I have made some changes. Please let me know what your thoughts are.

felipesere:
LGTM! Thanks!

lib/koans/10_map_sets.ex
```diff
@@ -5,19 +5,27 @@ defmodule MapSets do
...
   @set MapSet.new([1, 2, 3, 4, 5])
 
-  koan "I am merely another collection but, you can perform some operations on me" do
-    new_set = MapSet.new(@set, fn x -> 3 * x end)
-
-    assert MapSet.member?(new_set, 15) == ___
-    assert MapSet.member?(new_set, 1) == ___
-  end
+  koan "I am very similar to a list" do
+    assert Enum.fetch(@set, 0) == {:ok, ___}
```
iamvery: Might be nice to include an example here of fetching from the list _and_ the map.

```elixir
koan "Iamvery similar to a list" do
  list = [1,2,3]
  set = MapSet.new(list)

  assert Enum.fetch(list, 1) == 2
  assert Enum.fetch(set, 1) == 2
end
```

lib/koans/10_map_sets.ex
```diff
@@ -5,19 +5,27 @@ defmodule MapSets do
...
+  end 
 
   koan "However, I do not allow duplication" do
     new_set = MapSet.new([1, 1, 2, 3, 3, 3])
 
     assert MapSet.size(new_set) == ___
   end
 
+  koan "Even though I am like a list I am unordered after 32 elements" do
+    assert 1..33 |> MapSet.new() |> Enum.fetch(0) == {___, 11}
```
iamvery: I'm having a hard time grokking this lesson. What's going on here?

iamvery:
@msurekci thanks for being patient with me! Great work 😸

lib/koans/10_map_sets.ex
msurekci: MapSet is ordered only up until 32 elements because it is a Hash Mapped Array trie. So here I tried to demonstrate that it is an unordered collection.

iamvery: Oh wow, what a crazy little detail. I almost want to protect folks from this mess. Maybe it'd be better to just the point "sets do not guarantee order"

alexandru-calinoiu: Guys, I am doing the koans now and arrived at this exact koan (as you can guess I am new to elixir).

After I did not got any results from searching with various combinations of `elixir mapset 32 limit` I found this pr and your comments.

Can you please elaborate or point me to the elixir (or erlang) docs that explains this @lamvery's does not make sens with my limited knowledge.

alexandru-calinoiu: Never mind found it https://en.wikipedia.org/wiki/Hash_array_mapped_trie

Forgot about this from CS classes.

PR #124, opened 2016-05-26
Protocols

gemcfadyen:
@felipesere @ukutaht 

This introduces a koan for Protocols - demonstrated using Structs. This replaces PR https://github.com/elixirkoans/elixir-koans/pull/116

They show the positive case, and the case where you try to use a protocol but it is not supported. They also show the use of a default implementation using `@derive.`


felipesere:
LGTM. @ukutaht @Maikon any comments?

PR #127, opened 2016-06-10
Display inequality, guide learner

iamvery:
Show left and right hand values when equality fails
For assertions that contain function calls, displaying just the
expression is not helpful. For e.g.

    Assertion failed
    String.upcase("foo") == "Foo"

The learner is given no extra information to help them learn. Sure they
might consult the documentation (and should!), but perhaps a better
experience would be for us to inform the learner about the values it
encountered.

    Assertion failed
    String.upcase("foo") == "Foo"
    left:  "FOO"
    right: "Foo"

Learner: 💡 ah hah! it upcases the whole string! 💫

One might argue that this reveals too much information/makes it too easier. This has not been my experience in pairing with learners. What do you think?

### DIFF
--- a/lib/display/colours.ex
+++ b/lib/display/colours.ex
@@ -2,6 +2,7 @@ defmodule Display.Paint do
   def red(str), do: painter().red(str)
   def cyan(str), do: painter().cyan(str)
   def green(str), do: painter().green(str)
+  def yellow(str), do: painter.yellow(str)
 
   defp painter do
     case Mix.env do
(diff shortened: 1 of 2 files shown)


ukutaht:
Ha, I specifically didn't do this initially because I thought it would make the koans too obvious. The choice was arbitrary at the time and I'm glad that you've got some insights from pairing with actual learners :)

:+1: from me. @felipesere could you merge if you're happy as well?

iamvery:
I guess it depends on how folks use the koans. The problem that we ran into is "I have no insight into what that statement returns, I guess I'll experiment in IEx until I get it right." It feels better to iterate within the example itself. Like a test. Looking forward to @felipesere's feels :)

felipesere:
I personally like the current behaviour, but I've also seen beginners struggle with simply not knowing what to do next.

I'll merge it and consider adding a hard-more where the right side is not shown 😈

PR #128, opened 2016-06-10
Prefer function output over input

iamvery:
When learning about the behavior of a function, in general the learning
experience is better when provided with the input and asked for the
output.

This was felt while pairing on the koans with a learner. These particular examples felt slightly confusing as is for some reason. Of course, I do realize these aren't the only examples arranged in such a way, so we might consider rearranging them all. Alternatively, it might make sense to approach on a per example basis.


iamvery:
@felipesere any thoughts on this and applying it more generally throughout?

PR #129, opened 2016-06-11
Refactor error expansion to locate errors in koan module

iamvery:
The previous strategy wasn't robust enough to detect typos in function
calls. For example, if you misremembered String.duplicate/2 as
String.repeat/2 the koan runner would hang and have to be restarted.
This is because the stacktrace for an error like that doesn't have line
numbers. To avoid the problem, we detect the first pieces of the
stracktrace that is in the koan module. I can't decide if that's a
perfect solution, but it seemed to work in my testing. Hoping to get
other's feedback.

Before the change such an error would look like:

```
● HEAD ~/Code/OSS/elixir-koans » mix meditate --no-clear-screen
Compiled lib/options.ex
Compiled lib/runner.ex
Compiled lib/koans/01_equalities.ex
Welcome to the Elixir koans.
Let these be your first humble steps towards learning a new language.

The path laid in front of you is one of many.


Now meditate upon Equalities
|                              | 0 of 177
We shall contemplate truth by testing reality, via equality
Assertion failed in lib/koans/01_equalities.ex:12
true == ___



08:52:08.174 [error] Process #PID<0.139.0> raised an exception
** (FunctionClauseError) no function clause matching in Execute.extract_file_and_line/1
    (elixir_koans) lib/execute.ex:52: Execute.extract_file_and_line({Foo, :bar, [], []})
    (elixir_koans) lib/execute.ex:44: Execute.expand/1
    (elixir_koans) lib/execute.ex:35: Execute.exec/4
```

After the change:

```
● error-handling ~/Code/OSS/elixir-koans » mix meditate --no-clear-screen
Welcome to the Elixir koans.
Let these be your first humble steps towards learning a new language.

The path laid in front of you is one of many.


Now meditate upon Equalities
|                              | 0 of 177
We shall contemplate truth by testing reality, via equality
Assertion failed in lib/koans/01_equalities.ex:12
true == ___


Welcome to the Elixir koans.
Let these be your first humble steps towards learning a new language.

The path laid in front of you is one of many.


Now meditate upon Equalities
|                              | 0 of 177
We shall contemplate truth by testing reality, via equality
Error in lib/koans/01_equalities.ex:13
** (UndefinedFunctionError) undefined function Foo.bar/0 (module Foo is not available)
    (elixir) :elixir_def.__info__(:macros)
    (elixir) src/elixir_dispatch.erl:306: :elixir_dispatch.get_optional_macros/1
```

It stills a little awkward making changes to these pieces, because we don't have them under test. I didn't want to take time to figure that out at this point. It's frustrating for users to have to restart the runner anytime they make a small typo or misunderstand something.

### DIFF
--- a/lib/execute.ex
+++ b/lib/execute.ex
@@ -32,22 +32,21 @@ defmodule Execute do
 
   defp exec(module, name, args, parent) do
     result = apply(module, name, args)
-    send parent, expand(result)
+    send parent, expand(result, module)
     Process.exit(self(), :kill)
   end
 
-  defp expand(:ok), do: :ok
-  defp expand(error) do
+  defp expand(:ok, _), do: :ok
+  defp expand(error, module) do
     {file, line} = System.stacktrace
-      |> Enum.drop_while(&in_ex_unit?/1)
+      |> Enum.drop_while(&!in_koan?(&1, module))
       |> List.first
       |> extract_file_and_line
 
       %{error: error, file: file, line: line}
   end
 
-  defp in_ex_unit?({ExUnit.Assertions, _, _, _}), do: true
-  defp in_ex_unit?(_), do: false
+  defp in_koan?({module, _, _, _}, koan), do: module == koan
 
   defp extract_file_and_line({_, _, _, [file: file, line: line]}) do
    {file, line}


felipesere:
Thanks @iamvery for looking into this. 
I am a little busy this weekend, but I want to look at this a little closer. 

I think we can make the error less scary by possibly hiding the stack trace and show jus the top line. Maybe even a hint. What do you think?

I also want to bring more of this stuff under test.

iamvery:
@felipesere we're already absorbing errors and just showing them the file and line where it occurred. The problem was that for certain types of errors there was no file/line included in the stacktrace which caused the undefined function error. I'm not entirely convinced this is a perfect solution, but in my limited, manual testing it is "better" 😁

PR #134, opened 2016-06-15
A few small things I came across

iamvery:
No functional changes here. Just some tweaks.

### DIFF
--- a/lib/blanks.ex
+++ b/lib/blanks.ex
@@ -32,8 +32,8 @@ defmodule Blanks do
     |> elem(1)
   end
 
-  defp count(:___, acc), do: {node, acc+1}
-  defp count({:___, _, _}, acc), do: {node, acc+1}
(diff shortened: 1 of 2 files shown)


lib/blanks.ex
```diff
@@ -32,8 +32,8 @@ defmodule Blanks do
     |> elem(1)
   end
 
-  defp count(:___, acc), do: {node, acc+1}
-  defp count({:___, _, _}, acc), do: {node, acc+1}
```
felipesere: Good catch! How did this ever work? Where is node?

iamvery: Check out the commit notes https://github.com/elixirkoans/elixir-koans/commit/a2e3fe6d93b14e9105238f85554c8661e8ad9c53

felipesere: Thx!

PR #137, opened 2016-07-21
The koan about Mapset.member?/2 should appear before Mapset.member?/2 is used

Neku:


ukutaht:
That makes sense :+1:

There are some tests in `test/koans/map_sets_koans_test.exs` which are failing because you changed the order of the koans. All you need to do to make Travis happy is reorder those answers according to how they now appear.

ukutaht:
You may also want to add yourself to `CONTRIBUTORS.md` 😉

ukutaht:
Awesome, thank you @Neku ! ❤️

PR #142, opened 2016-10-12
Changed the quotations from singles to doubles to keep consistency in…

jmphilips:
… numbers_koans_test.exs

### DIFF
--- a/test/koans/numbers_koans_test.exs
+++ b/test/koans/numbers_koans_test.exs
@@ -8,7 +8,7 @@ defmodule NumbersTests do
       true,
       [5, 8, 1, 2, 7],
       1234,
-      '7',
+      "7",
       "1234",
       42,
       " years",


iamvery:
@jmphilips the correct answer must be single quoted, because it's a charlist (see https://github.com/elixirkoans/elixir-koans/blob/9ee83336d6d57e6804ef0ae74c1cc2c6b56ff909/lib/koans/04_numbers.ex#L27)

Does that make sense?

jmphilips:
Yeah, totally.  Really sorry!  I am obviously just getting into Elixir, and
I was really looking forward to making my first "open source contribution",
and I guess I got a little carried away.  Thank you so much for keeping up
with the elixir koans.  I look forward to diving deeper into the koans and
elixir in the future.

iamvery:
No problem at all! Thanks for thinking critically about this stuff. That's how it gets better. Unfortunately I don't have commit-bits here, so either you can close or we can wait on someone that does 😊

PR #146, opened 2016-10-17
A few tweaks to Blanks

iamvery:
Just a few small clean ups. The only thing I'm not completely confident with is the removal of the `:do` function heads. I suppose this works because the Macro.prewalk used further down the call stack can account for it?

### DIFF
--- a/lib/blanks.ex
+++ b/lib/blanks.ex
@@ -11,11 +11,9 @@ defmodule Blanks do
     {args, replacements} = Macro.prewalk(args, replacements, &pre_pin/2)
     {put_elem(node, 2, args), replacements}
   end
-  defp pre(:___, [first | remainder]), do: {first, remainder}
   defp pre({:___, _, _}, [first | remainder]), do: {first, remainder}
   defp pre(node, acc), do: {node, acc}
 
-  defp pre_pin(:___, [first | remainder]), do: {pin(first), remainder}
   defp pre_pin({:___, _, _}, [first | remainder]), do: {pin(first), remainder}
   defp pre_pin(node, acc), do: {node, acc}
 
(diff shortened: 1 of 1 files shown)


ukutaht:
Much cleaner 👍 

When you say you're not confident in removing the `:do` function heads, are you worried about it blowing up in some cases? Or that it works fine and you're not sure why? Just want to make sure we're not merging something that might break the koans.

iamvery:
Yeah, I guess I'm concerned about there being a case that's not covered by the tests. I _think_ it's not a problem because the call to Macro.prewalk further down the call stack can handle expressions wrapped in the keyword list, i.e. `[do: expr]`. It's just hard to submit it with absolute confidence since I've not had a failing test or specific case illustrating the need for it. Is that fair? FWIW, I did worth through a few of the koans manually and didn't have any problems. I have a suspicion that function head was naively introduced under an assumption it would be required.

felipesere:
Yeah this looks much cleaner.
Would it make sense to rename the `replacements` that gets assigned in the body? 
`foo = something(foo)` feels a little iffy...

iamvery:
I guess I'm generally OK with reassignment when it happens right at the top of the functions. Seems pretty clear to me that it's making sure the value is a list. The plurality of the variable also helps communicate this. I suppose if we really wanted to change the variable, we could use `replacements_list` as the second one? What do you think?

felipesere:
How about inlining `List.wrap(replacements)`?
I looked at it again and you are right, its fairly obvious what happens with `foo = bar(foo)`

iamvery:
I guess I still prefer to be able to think of `replacements` as a list throughout the function. I'm not a fan of the way this reads either:

```elixir
ast
  |> Macro.prewalk(List.wrap(replacements), &pre/2)
  |> elem(0)
```

I'm happy to change if you're strongly leaning that way :). Otherwise :shipit:

felipesere:
How about you merge it yourself? You should have access by now 😺

iamvery:
😍

PR #148, opened 2016-10-24
Bump elixir version.

marosluuce:
This will require the version to be at least 1.3 and less than 1.4. I'm assuming that most people using this aren't bound to a specific, older version and will be able to update. We'll have to revisit this when 1.4 comes out, but at that point we'd probably want to update the travis build anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: elixir
-elixir:
-  - 1.2.2
-otp_release:
-  - 18.2.1
+matrix:
+  include:
+    - elixir: 1.2.2
+      otp_release: 18.2.1
+    - elixir: 1.3.4
+      otp_release: 19.1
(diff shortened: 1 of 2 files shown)


iamvery:
Do we want to be so restrictive on the Elixir version folks can use? I do hope people learning are using a modern version, but if it works on something older or newer it seems fine to allow that. If the project does stay open to a wider version range, perhaps a build matrix should be set up on CI to confirm the build on various versions. Thoughts?

marosluuce:
I'm not opposed to supporting older versions. I would like to avoid many test runs on CI though. Maybe the compromise is to test with the oldest supported version and the newest version?

iamvery:
That seems fair 🙂

felipesere:
This looks good. I'd be happy to take versions on/off the matrix as people report issues. 
In general, I think the bracket between lowest and highest supported version should be enough.

PR #150, opened 2016-12-23
Format match failures

### LINKED ISSUE #144: Handle unary failures more elegantly
Currently failures for assertions such as `assert match?(1,2)` are displayed as if they're being made on binary statements. Here's an example:

```
Now meditate upon PatternMatching
|==>                           | 7 of 73
The pattern can make assertions about what it expects
Assertion failed in lib/koans/12_pattern_matching.ex:49
match?([1, _second, _third], [1])

left:  :ex_unit_no_meaningful_value
right: [1]
```

This unfortunately leaks ExUnit's `:ex_unit_no_meaningful_value` detail through and causes confusion (see #143) to learners.

I believe the solution would be to handle these so-called "unary" (i.e. one argument) assertions different than so-called "binary" assertions.

If I have some time to look at it I will. Just reporting this issue in case it goes to someone else ❤️

iamvery:
This formats match failures to provide clearer learner feedback. ExUnit has a [special case](https://github.com/elixir-lang/elixir/blob/3b499ffb87269d3fa7c874e1deb55b3a30435027/lib/ex_unit/lib/ex_unit/assertions.ex#L145-L158) for `match?`. Also, I went ahead and added tests around inequality failures and it caught a missing interface on the test painter! 💫 

screen shot 2016-12-23 at 8 10 37 am

[Closes #144]

### DIFF
--- a/test/display/failure_test.exs
+++ b/test/display/failure_test.exs
@@ -5,13 +5,13 @@ defmodule FailureTests do
   test "assertion failure with proper expression" do
     error = error(%ExUnit.AssertionError{expr: "hi"})
 
-   assert Failure.format_failure(error) == "Assertion failed in some_file.ex:42\n\"hi\"\n"
+    assert Failure.format_failure(error) == "Assertion failed in some_file.ex:42\n\"hi\"\n"
   end
 
   test "assertion failure with message" do
     error = error(%ExUnit.AssertionError{expr: :ex_unit_no_meaningful_value, message: "hola"})
 
-   assert Failure.format_failure(error) == "Assertion failed in some_file.ex:42\nhola\n"
+    assert Failure.format_failure(error) == "Assertion failed in some_file.ex:42\nhola\n"
   end
 
   defp error(error) do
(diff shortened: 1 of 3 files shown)


test/display/failure_test.exs
```diff
@@ -5,13 +5,36 @@ defmodule FailureTests do
...
 
-   assert Failure.format_failure(error) == "Assertion failed in some_file.ex:42\n\"hi\"\n"
+    assert Failure.format_failure(error) == "Assertion failed in some_file.ex:42\n\"hi\"\n"
   end
 
   test "assertion failure with message" do
     error = error(%ExUnit.AssertionError{expr: :ex_unit_no_meaningful_value, message: "hola"})
 
-   assert Failure.format_failure(error) == "Assertion failed in some_file.ex:42\nhola\n"
+    assert Failure.format_failure(error) == "Assertion failed in some_file.ex:42\nhola\n"
```
felipesere: I still recognize my Spanish roots in this assertion 😸

Issue #153, opened 2016-12-23
Come up with better strategy for testing "do unwrapping"


iamvery:
See #152 

https://twitter.com/felipesere/status/812324732850892801

Issue #151, opened 2016-12-23
Progress display messed up due to recent change


iamvery:
I accidentally introduced a change that messed up the progress display a little while back. I've got a fix coming up, but I'm creating this issue so that I can link to something with pictures in the commit notes.

[The problematic change](https://github.com/elixirkoans/elixir-koans/pull/146/files#diff-37c7c40e84e68a928573e3db6fe872e7L39) failures to unwrap the koan body from it the keyword list argument, and therefore the display cannot be formatted correctly while running the koans.

Currently, it looks like this at it worst:
![screen shot 2016-12-23 at 8 52 22 am](https://cloud.githubusercontent.com/assets/79619/21456371/30f22550-c8ed-11e6-9fe2-d7f459bf0557.png)

However, only the offending line is supposed to be displayed:
![screen shot 2016-12-23 at 8 52 35 am](https://cloud.githubusercontent.com/assets/79619/21456374/3926cd48-c8ed-11e6-937b-6d61aa96a27c.png)

Fix incoming...

PR #154, opened 2016-12-23
Test for single arity failures

felipesere:
This is the best that could come up with to deal with #153 

It does fail when the `do: ` is missing, but it is not immediately obvious...

@iamvery what do you think? Something to pivot on?

### DIFF
--- a/test/display/failure_test.exs
+++ b/test/display/failure_test.exs
@@ -37,7 +37,7 @@ defmodule FailureTests do
                                             """
   end
 
-  test "single arity failure" do
+  test "only offending lines are displayed for errors" do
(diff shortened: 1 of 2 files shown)


iamvery:
The test works! But I think the description may be a little off. It's less about single-arity koans and more about displaying only the offending line in failures. The description and screenshots in #151 illustrate it pretty well. Without unwrapping the body from to `:do` key, Blanks.replace_line/2 can't properly traverse the expression. Am I making any sense? I'm finding this hard to get my head around for explanation..

iamvery:
I think it's better! And better yet, it covers the problem well so ima :shipit:

Issue #138, opened 2016-08-23
mix meditate doesn't work with intellij


jkbbwr:
After closing the project, installing `inotify-tools` and running `mix meditate` it doesn't detect any file changes from intellij

cliche818:
I had to actually turn off safe write in IntelliJ (Appearance & Behavior > System Settings > Synchronization > check off Use "safe write".  

My fix didn't really help as after a few edits, the watcher will fail to reload the right file.

It seems IntelliJ write temp files like
/elixir-koans/lib/koans/03_sigils.ex___jb_old___
/elixir-koans/lib/koans/03_sigils.ex___jb_tmp___

And it writes multiple times when I edit a file which I think causes the watcher to not keep up.

jkbbwr:
This is not uncommon. Is there a way this can be tracked and fixed.

jkbbwr:
Thanks very much!

leapingfrogs:
@cliche818 - I'm not experiencing the issue you mention in not keeping up with edits, but I'm curious whether that was impacted by reloading for all files not just the .ex files.  Would filtering the jb_tmp file suffix as in #156 help at all?

iamvery:
Looks like #156 fixes the reload detection problem in IntelliJ. I guess we'll give @jkbbwr a couple days to confirm, and we'll close out this issue. Thanks everyone! ❤️

cliche818:
@leapingfrogs Doesn't seem to help, I am using the latest version of IntelliJ with the save write checked to try this out.  Same thing happens as before.  It does reload a few times, but then it dies.  Sometimes, it even says Code.LoadError could not load [the file I just changed].

iamvery:
@cliche818 you mentioned above both "jb_tmp" and "jb_old" files being generated. The "normalize" solution only accounts for "tmp" files. Do you think that might be related?

jkbbwr:
yea reload aint working

felipesere:
@jkbbwr @cliche818 can you tell me a bit more about your setup?
I am running `mix meditate` in iTerm and editing the files in IntelliJ and it seems happy.
Oh and all of this is in OSX.

I just pushed an update the helped the reload in Ubuntu. Maybe that solved it?

jkbbwr:
Seems to magically fix it for me.
`Linux devbox2 4.9.8-1-ARCH #1 SMP PREEMPT Mon Feb 6 12:59:40 CET 2017 x86_64 GNU/Linux`
[...]

PR #156, opened 2017-01-19
Normalize the file before calling reload.

leapingfrogs:
Fixing: https://github.com/elixirkoans/elixir-koans/issues/138

By normalizing first we remove the IntelliJ specific suffix if it is
present which allows the ".ex" extension to correctly match IntelliJ
edited files.

### DIFF
--- a/lib/watcher.ex
+++ b/lib/watcher.ex
@@ -3,7 +3,7 @@ defmodule Watcher do
 
   def callback(file, events)  do
     if Enum.member?(events, :modified) do
-      reload(file)
+      file |> normalize |> reload
 
       if Tracker.complete? do
         Display.congratulate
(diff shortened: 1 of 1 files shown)


iamvery:
This makes sense to me. Basically in order for [`Path.extname(file) == ".ex"`](https://github.com/elixirkoans/elixir-koans/pull/156/files#diff-517dd893f267577b9a361cbbb8770363R16) to be true, the file must first be `normalized`. Is that what you're thinking?

felipesere:
Looks good to me.

leapingfrogs:
@iamvery, yes that's correct.

iamvery:
I haven't ever used IntelliJ, but I installed it to see this issue in action. On master, editing a file does not trigger re-run. On your branch, it does! That's good enough for me 😄 . Thanks for helping improve things! ❤️

PR #160, opened 2017-01-22
Also test for Elixir 1.4.0

felipesere:


iamvery:
I was thinking a little more about what Elixir versions to target and test for this project. I think in the past I was sort of in the mindset of ALL the versions. Now I'm leaning towards only the latest. I believe this makes sense as the target audience is new to the language and likely to have installed the latest from the website or whatever. Anyway just leaving my thoughts here :)

Issue #161, opened 2017-01-22
Nested Dictionary Structures


srufle:
I am working my way through "Programming Elixir 1.3"
Chapter 8 - Nested Dictionary Structures
http://bit.ly/2kgQ5si

Uses get_in, put_in, update_in, and get_and_update_in. Should I modify one of the existing koans [maps, map_sets, structs]? or create a new one.

felipesere:
Welcome!
I'd suggest extending the existing koan. Make sure to also fill in the right test 😄

srufle:
Thx, 
How does maps_koans_test.exs relate to 09_maps.ex
How do I run tests.

felipesere:
`09_maps.ex` are the actual koans that the user will run.
`maps_koans_test.exs` are the answers to the koans and a function to feed the answers - in the same order! - to the koans.

So if you add a new koan somewhere in the middle, you'll have to add the answer in the same position in the `answers` list.

You can run all tests by running `mix test`

PR #162, opened 2017-01-22
Demonstrate put_in to manipulate nested struct values

srufle:
#161

### DIFF
--- a/lib/koans/11_structs.ex
+++ b/lib/koans/11_structs.ex
@@ -59,4 +59,14 @@ defmodule Structs do
     airline = %Airline{plane: %Plane{maker: :boeing, passengers: 200}}
     assert update_in(airline.plane.passengers, &(&1 + 2)) == ___
   end
+
+  koan "Use the put_in macro with atoms to replace a nested value in a non-struct" do
+    airline = %{
+                  plane: %{
+                    maker: :boeing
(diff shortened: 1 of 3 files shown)


felipesere:
Are we covering the most common cases of put_in and friends? I haven't used them before. 
Also, add yourself to the CONTRIBUTORS.md 🙂

lib/koans/11_structs.ex
```diff
@@ -59,4 +59,14 @@ defmodule Structs do
     airline = %Airline{plane: %Plane{maker: :boeing, passengers: 200}}
     assert update_in(airline.plane.passengers, &(&1 + 2)) == ___
   end
+
+  koan "Use the put_in macro with atoms to replace a nested value in a non-struct" do
+    airline = %{
```
iamvery: Maybe one-line this map like the structs in the above tests?

srufle: K, I will put it all on one line. I make things multiline so I can see teh brace matches easier.

iamvery: That makes sense to me! I was just thinking it should be consistent across tests (either one line or multi line)

felipesere:
Sorry I forgot to merge this. Thank you so much! ❤️

PR #163, opened 2017-01-24
should call anonymous function with dot

kaddopur:
otherwise the `greet/1` will be called
https://github.com/elixirkoans/elixir-koans/blob/master/lib/koans/13_functions.ex#L6-L8

@dillonkearns


iamvery:
Good catch! I went ahead and [changed up](b8d7abe4ea11512f5ba38d0f9d4b65779ba8cb23) the anonymous function a little too so the behavior is distinct. Thanks for you help! ❤️

PR #165, opened 2017-01-29
Support for Elixir 1.4.1

rafax:


mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Koans.Mixfile do
   def project do
     [app: :elixir_koans,
      version: "0.0.1",
-     elixir: ">= 1.2.1 and <= 1.4.0",
+     elixir: ">= 1.2.1 and <= 1.4.1",
```
iamvery: I wonder if we should leave this more open ended to lower thrash in the config @felipesere @ukutaht?

ukutaht: Good point.  I don't know semver requirement syntax very well so here's the best I could come up with: `~> 1.2.1 or ~> 1.3.0 or ~> 1.4.0`

This would allow everything from `1.2.1`, through the `1.3` and `1.4`  minor releases regardless of the patch version as you can see:

screen shot 2017-01-30 at 15 57 08

_Off topic: how cool is that `Version` module?_

I remember something breaking on `1.2.0` which is why the low end cares about the patch version.

felipesere: I like Ukus suggestion. I think we can even go with `~1.2.1 or ~> 1.3 or ~1.4`

iamvery: I like the idea of `~>1.4` on the upper end, which I _think_ means "anything greater than or equal to `1.4` up to, but not including `2.0`. That seems like a good target for us to not have to tweak this every time.

Something else to consider, as the koans grow, we're likely to have examples for features of the language not found in earlier versions. So it's probably a good idea to have both the lower and upper bounds of the version range in CI just to try and ensure the examples are valid for folks using versions in the range.

ukutaht: @iamvery `~> 1.4` basically means `1.4.x`. 

For `1.5` we would have to update the config still. I think what you're suggesting is something like `>= 1.2.1 and < 2.0.0`. If Elixir follows semver properly, there shouldn't be any breaking changes until 2.0 so I'd be fine with that as well.

iamvery: I have a hard time keeping the semantics of `~>` in my head, but I _think_ what you said is incorrect. I believe `~>1.4` is basically `1.x.y` where `x` >= `4`.

```
iex(1)> Version.match?("1.5.0", "~>1.4")
true
```

am i seeing that right?

felipesere: Can I merge this? 😄

iamvery: We can continue the discussion in #167

PR #170, opened 2017-02-11
Clarify the koan that shows that MapSets are unordered.

### LINKED ISSUE #169: The "unordered MapSet" koan is very unclear
The [koan](https://github.com/elixirkoans/elixir-koans/blob/master/lib/koans/10_map_sets.ex#L18-L20) that shows that MapSets aren't ordered is confusing:
```
koan "Even though I am like a list I am unordered after 32 elements" do
    assert 1..33 |> MapSet.new() |> Enum.fetch(0) == {___, 11}
end
```

There are two reasons:

1. "*after* 32 elements" is grammatically misleading and the number 32 is unexplained.
1. It's unclear to beginners where the number 11 comes from.  For those who understand where the 11 could come from (some incidental determinism?), it seems that this koan is brittle as there's no guarantee that the runtime would put this item in the beginning (I assume).

huyz:
Fixing #169 

[Closes #169]

### DIFF
--- a/lib/koans/10_map_sets.ex
+++ b/lib/koans/10_map_sets.ex
@@ -15,8 +15,11 @@ defmodule MapSets do
     assert MapSet.size(new_set) == ___
   end
 
-  koan "Even though I am like a list I am unordered after 32 elements" do
-    assert 1..33 |> MapSet.new() |> Enum.fetch(0) == {___, 11}
+  koan "Even though I am like a list I am unordered, especially with 33+ elements" do
+    new_set = MapSet.new 1..33
+    ordered_list = Enum.to_list 1..33
+    assert (ordered_list == MapSet.to_list(new_set)) == __
+    assert (ordered_list == MapSet.to_list(new_set) |> Enum.sort()) == __
   end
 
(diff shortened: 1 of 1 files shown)


huyz:
Hmm, don't know what the CI tests fail.  What could I have broken to get the error `no environment variables set`.

lib/koans/10_map_sets.ex
```diff
@@ -15,8 +15,11 @@ defmodule MapSets do
     assert MapSet.size(new_set) == ___
   end
 
-  koan "Even though I am like a list I am unordered after 32 elements" do
-    assert 1..33 |> MapSet.new() |> Enum.fetch(0) == {___, 11}
+  koan "Even though I am like a list I am unordered, especially with 33+ elements" do
```
iamvery: What about "Event though I seem to act like a list, ordering isn't guaranteed, especially with more than 32 elements"?

iamvery: As you mention in #169, the number "32" is unexplained. TBH I don't remember what's special about that either. I wonder if a comment close by with more explanation might be interesting to have?

Issue #166, opened 2017-02-02
about koan "New processes are spawned functions"


Airmacho:
I am a beginner of elixir learning. Thank you for this.

As to koan "New processes are spawned functions" in [15_processes.ex](https://github.com/elixirkoans/elixir-koans/blob/master/lib/koans/15_processes.ex#L20) file.

My iex(elixir 1.4.1) returns `false` when I use Process.alive?(pid)

felipesere:
Thank you for reporting this issue!
I just updated to 1.4.1 and tried it myself and got through it normally.
Can you try 'mix test' which will run all the Koans with the appropriate answers?
If that fails too, we'll have to dig deeper...

iamvery:
It seems the test expects the answer to be true (see bef1cf), but I also observed that it's false in the console. I'm pretty sure this is a timing problem. Since nothing in the spawn process is blocking, it's only alive for a _very short_ amount of time. The tests run really fast, so it gets to the alive check before the spawned process dies.

```elixir
iex(1)> pid = spawn(fn -> nil end); Process.alive?(pid)
true
iex(2)> pid = spawn(fn -> nil end)
#PID<0.88.0>
iex(3)> Process.alive?(pid)
false
```

So in general, this is a bad/confusing example, and we should consider updating it.

felipesere:
Yeah, I was thinking the same and started experimenting with that on my laptop.
We could put a recursive function in there that will loop forever?

iamvery:
Perhaps a `receive` call would be more idiomatic?

felipesere:
Looks good 😄

jessejanderson:
This seems backward to me... shouldn't `Process.alive?` return false to indicate that when a process has nothing to do, it dies immediately?

The `receive` example seems like a hack to me in the current context that teaches false information.

Maybe this should be split into 2 tests. The first is as written (maybe with a tweaked description) expecting an answer of `false`, then the next test has a receive and can return true?

jessejanderson:
(maybe I'm just annoyed that I knew the process would be dead immediately, and the test was telling me that I was wrong) :D

iamvery:
@jessejanderson that's a good point. To me it basically feels like "concurrency is hard. look concurrency is hard to teach too" 😆 

[...]

PR #173, opened 2017-02-18
Just 'nil' might have lived long enough, 'receive' definitly does.

felipesere:
Should clarify #166 

`receive do; end` blocks the spawned thread so it's guaranteed to be alive when we assert.

### DIFF
--- a/lib/koans/15_processes.ex
+++ b/lib/koans/15_processes.ex
@@ -18,7 +18,9 @@ defmodule Processes do
   end
 
   koan "New processes are spawned functions" do
-    value = spawn(fn -> nil end)
+    value = spawn(fn -> receive do
+                      end
+                end)
 
     assert is_pid(value) == ___
   end


lib/koans/15_processes.ex
```diff
@@ -18,7 +18,9 @@ defmodule Processes do
   end
 
   koan "New processes are spawned functions" do
-    pid = spawn(fn -> nil end)
+    pid = spawn(fn -> receive do
+                      end
+                end)
```
iamvery: Thanks for going ahead with at least fixing this problem! (Despite the examples themselves may needed to be clearer). Do you have any specific opinions about the style of this? The way it is right now is a little 😖 . What do you think about naming the function?

```elixir
wait_forever = fn ->
  receive do
  end
end
pid = spawn(wait_forever)
```

Maybe that would help make things a little clearer in the mean time? I had also used the form `receive(do: nil)` in #166, but I'm not sure how confusing that is.

PR #175, opened 2017-02-19
String.strip is deprecated, replace it with String.trim

Nagasaki45:


iamvery:
@felipesere this happened [in 1.3](https://github.com/elixir-lang/elixir/blob/v1.3/CHANGELOG.md#3-soft-deprecations-no-warnings-emitted). However, our build still includes 1.2 (thus the test failure). Thoughts?

felipesere:
Is there a way to get numbers from the community about version usage? I'd like to cover 70% of Elixir usage (number pulled out of thin air)

iamvery:
@felipesere i'm not sure. I wonder if https://hex.pm/ would have any insight?

felipesere:
This can be merged once we merge #183

felipesere:
@Nagasaki45 can you rebase `master` into this to ensure CI still passes? We have dropped support for Elixir 1.2 so this should work now.

Nagasaki45:
@felipesere, done! anything else?

PR #180, opened 2017-03-22
Split koan with race condition into two more logical ones

cybrox:
I just discovered this project today and decided to work my way through it out of curiosity if I could _solve_ it all. Apparently, I am not the only one confused by the _New processes are spawned functions_ koan, so this is my approach to solving the issue.

This was previously discussed in #166 and PR'd #173, however, this seems like the most logical way of putting it from a teaching perspective. First establishing that spawned processes behave like our main process, using the same way of asserting them, and then adding the (imo quite important) information about process lifetime.


lib/koans/15_processes.ex
```diff
@@ -17,9 +17,20 @@ defmodule Processes do
     assert is_pid(self()) == ___
   end
 
-  koan "New processes are spawned functions" do
+  koan "Like our current process, new processes are spawned functions" do
     pid = spawn(fn -> nil end)
 
+    assert is_pid(pid) == ___
```
iamvery: I wonder if it would help to really drive that point home if you included two assertions here:

```elixir
assert is_pid(self) == true
assert is_pid(value) == ___
```

You might also rename the return value of the call to spawn/1 to something more generic as to not give it away. What do you think?

lib/koans/15_processes.ex
```diff
@@ -17,9 +17,20 @@ defmodule Processes do
...
+  end
+
+  koan "Processes exit after executing their function, unless they wait for something" do
```
iamvery: This might be clearer if the condition lead the statement.

This lesson still bugs me though. Ideally you'd want the learning process to be something like:
1. processes are spawned functions
2. processes die when the function exits
3. processes can send messages
4. a process waiting for a message will never die

Does that seem clearer? The trick is #2 is really hard. That's the issue we have with the current lesson, specifically that:

```
pid = spawn(fn -> nil end)
assert Process.alive?(pid)
```

should not pass intuitively, but it does because the `assert` executes in the current process before the spawned process exist. concurrency щ（ﾟДﾟщ）

lib/koans/15_processes.ex
iamvery: Nevermind my recommendation about `is_pid(self)`. I read the lesson in context and see that it's covered just before this.

iamvery:
With your inspiration, I took another shot at this. Went pragmatic and wrote a little note to explain the mysterious timer.

What do you think?

```elixir
  koan "New processes are spawned functions" do
    value = spawn(fn -> nil end)

    assert is_pid(value) == true
  end

  koan "Processes die when their function exits" do
    pid = spawn(fn -> nil end)

    # The spawned function executes concurrently with the current process. So
    # you must allow just a little bit of time to pass in order to know
    # reliably that it has completed executing and exited. 10ms should do it.
    :timer.sleep(10)

    assert Process.alive?(pid) == false
  end

  koan "Processes can send and receive messages" do
    send self(), "hola!"

    receive do
      msg -> assert msg == "hola!"
    end
  end

  koan "A process will wait forever for a message" do
    wait_forever = fn ->
      receive do
      end
    end

    pid = spawn(wait_forever)

    assert Process.alive?(pid) == true
  end
```

cybrox:
Thanks for the comments!

I was playing with the idea of just putting a timer in as well but I kind of got obsessed with the idea of writing self explanatory code. Guess that's not exactly easy for explaining concurrency, though.

I do very much like your pragmatic example. Maybe even going overboard and having an additional _long running process_ to really show the difference would make things even clearer?

```
  koan "Processes die when their function exits" do
    fast_process = spawn(fn -> :timer.sleep(10) end)
    slow_process = spawn(fn -> :timer.sleep(1000) end)

    # All spawned functions are executed concurrently with the current process.
    # You check back on slow_process and fast_process 50ms later. Let's
    # see if they are still alive!
    :timer.sleep(50)

    assert Process.alive?(fast_process) == false
    assert Process.alive?(slow_process) == true
  end
```

iamvery:
I like it 😊. what do you think @felipesere @ukutaht?

felipesere:
I might be able to take a look at it tonight.
Thanks for the effort thought! 🙂

cybrox:
👍  I added the code and tests for a combination of @iamvery's example and my changes that we discussed so it can be reviewed.

One last thing I noticed while adding that. It might make sense to move the koan _Waiting for a message can get boring_ up to be brought up after the wait forever example. Or to extend the list you commented a bit:

1. processes are spawned functions
2. processes die when the function exits
3. processes can send messages
4. a process waiting for a message will never die
5. ...but it can report when it has been waiting for too long

iamvery:
Looking great @cybrox! I like the idea about moving it up and/or rewording ✨

PR #181, opened 2017-03-24
Add Koans for GenServer

cybrox:
I originally wanted to take on #46 but figured GenServers would be the most important piece missing between Tasks, Agents and eventually Supervisors. 

This is pretty much a draft, feedback is welcome 😄


felipesere:
It seems like GenServer is a bit picky about Elixir 1.2
If personally feel that I'd sacrifice Elixir 1.2 support in favour of GenServer koans.
@iamvery @ukutaht what do you think?

iamvery:
same problem as #175

cybrox:
Oh, pardon, I only ran the tests on the latest version ( `1.4.1`), didn't test with `1.3.x` and `1.2.x`

Personally, since `elixir-koans` is directed towards people starting to use elixir, I think it does make sense to promote the idea of using the latest version over older ones. However, let me know about your decision when it comes to 1.2 support. If you're in favour of keeping support, I will see if I can get it to work on 1.2 as well.

felipesere:
Elixir 1.3 was released in June 2016. I think we are safe to remove it by now.
I wonder if we can emit a warning as part of `mix.exs`?

felipesere:
Once #183 lands we can merge this too.

felipesere:
@cybrox If you rebase `master` into this we can merge it.

cybrox:
@felipesere Sorry, I completely forgot that I still wanted to test this with `1.2`.    
I rebased it now and tests without `1.2` look green 👍

lib/koans/18_genservers.ex
```diff
@@ -0,0 +1,68 @@
...
+  koan "You can think of GenServers like more flexible agents" do
+    {:ok, pid} = GenServer.start_link(ExampleGenServer, :ok, [])
+
+    assert is_pid(pid) == ___
+  end
+
+  koan "GenServers can have a name too" do
+    {:ok, pid} = GenServer.start_link(ExampleGenServer, :ok, name: StackServer)
+
+    assert is_pid(GenServer.whereis(StackServer)) == ___
```
iamvery: Would it be worth showing that the result of `GenServer.whereis(StackServer) == pid` maybe something like

```elixir
{:ok, pid} = GenServer.start_link(ExampleGenServer, :ok, name: StackServer)
found_pid = GenServer.whereis(StackServer)
assert pid == found_pid
```

Not sure how to set that up exactly off the top of my head, but does it at least make sense?

lib/koans/18_genservers.ex
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  koan "Use cast to send an asynchronous request to the server" do
+    GenServer.start_link(ExampleGenServer, :ok, name: StackServer)
+
+    GenServer.cast(StackServer, {:push, 1})
+    GenServer.cast(StackServer, {:push, 3})
+    GenServer.cast(StackServer, {:push, 5})
+
+    :timer.sleep(10)
```
iamvery: Since the requests are async, you need this to allow enough time to pass that all messages are handled, yes?

lib/koans/18_genservers.ex
```diff
@@ -0,0 +1,68 @@
...
+    stack = GenServer.call(StackServer, :peek)
+
+    assert stack == ___
+  end
+
+  koan "Casts always return :ok, they can't be trusted" do
```
iamvery: Perhaps rather than "can't be trusted", we should say something like "the return value is not meaningful"

lib/koans/18_genservers.ex
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    # Use casts to perform asynchronous actions
+    def handle_cast({:push, value}, stack) do
+      {:noreply, [value|stack]}
+    end
+  end
+
+
+  koan "You can think of GenServers like more flexible agents" do
```
iamvery: It's hard to see by the examples how GenServers are "more flexible" than Agents. Part of what is useful about GenServers is the common interface, right? I'm pulling a little bit from memory here, so forgive errors, but the interface is used by supervisors, yes?

lib/koans/18_genservers.ex
felipesere: I like the above 👍

lib/koans/18_genservers.ex
```diff
@@ -0,0 +1,68 @@
...
+
+    assert stack == ___
+  end
+
+  koan "Use cast to send an asynchronous request to the server" do
+    GenServer.start_link(ExampleGenServer, :ok, name: StackServer)
+
+    GenServer.cast(StackServer, {:push, 1})
+    GenServer.cast(StackServer, {:push, 3})
+    GenServer.cast(StackServer, {:push, 5})
```
felipesere: Wouldn't the same example work with `call`?
Why would I choose one over the other? What is an interesting case for `sync` over `async`?

felipesere:
Is it worth while creating more servers that emphasize the cases where `call` is advantageous and the ones where `cast` is?

Nagasaki45:
This is really nice addition to the koans!
As a relatively new elixir dev I would like to add some comments:
- The use of `pid`s with gen servers is not shown. What do you think about using all of the examples with `pid`s and introducing named gen servers only in the end (after discussing `cast` and `call`)?
- Naming gen servers is a great feature, and it's quite important to know when you start learning OTP. However, I think that `GenServer.whereis` is less important and not crucial for appreciating the ability to name gen servers. I would prefer not to show it on this stage, and especially not in the first demo of named gen servers.
- The examples doesn't use the `pop` message implemented by the `ExampleGenServer`. I believe that a better demo of the functionallity of this gen server will be with only `push`s and `pop`s instead of the `peek` used. For example, instead of `push`ing 3 elements and then `peek`ing to check the result we can `pop` and check 3 elements one by one.

lib/koans/18_genservers.ex
cybrox: As far I remember, I don't think the interface is actually _used_ by supervisors, however, their interfaces especially in terms of initialisation have matching specs, which makes them easy to handle together.

I do agree that the comparison to agents is probably not the best idea, though. I wanted to hint previous koan-sets with that but agree that there's way better comparisons.

lib/koans/18_genservers.ex
cybrox: 👍  for that as well. I will add/refactor some koans to make handling more pid related, as @Nagasaki45 mentioned as well.

lib/koans/18_genservers.ex
cybrox: Yep, it looks very unclean, though. I might have a better idea though, see below.

lib/koans/18_genservers.ex
cybrox: It would, yes. I agree that the demonstration of `sync`/`async` here is a bit sparse.
I will think of an example similar to `fast_process` and `slow_process` (but wrapped in a GenServer) from #180, with that, I should also get rid of that possibly confusing `:timer.sleep(10)`

felipesere:
@cybrox We have just merge another PR that added the GenServer koans. 
Thank you for all the effort you put into this PR.
I highly encourage you to open a pull request that enhances the current GenServer koan if you have ideas that have not been implemented.

PR #182, opened 2017-03-30
Add Streams

### LINKED ISSUE #12: Koans about Enumerables and Streams

fabientownsend:
This pull request add some basic koan about stream reading

[Closes #12]

### DIFF
--- a/lib/koans/19_streams.ex
+++ b/lib/koans/19_streams.ex
@@ -1,0 +1,27 @@
+defmodule Streams do
+  use Koans
+
+  @intro "Streams"
+
+  koan "Streams are lazier than Enum, they only return a struct" do
+    assert Stream.map([1, 2, 3], &(&1 + 1)).__struct__ == ___
+  end
+
+  koan "You have to ask it explicitly to do the operation" do
+    updated_list =
+      [1, 2, 3]
+      |> Stream.map(&(&1 + 1))
+      |> Enum.to_list
+
+    assert updated_list == ___
+  end
+
+  koan "Which will let you decide which element you want to compute in the stream" do
+    computed_element =
+      1..42
+      |> Stream.map(&("I computed: #{&1}"))
+      |> Enum.take(1)
+
+    assert computed_element == ___
+  end
+end
--- a/lib/koans/19_streams.ex
+++ b/lib/koans/19_streams.ex
@@ -7,7 +7,7 @@ defmodule Streams do
     assert Stream.map([1, 2, 3], &(&1 + 1)).__struct__ == ___
   end
 
-  koan "You have to ask it explicitly to do the operation" do
+  koan "Streams are executed by terminal operations" do
     updated_list =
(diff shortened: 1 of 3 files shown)


lib/koans/19_streams.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Streams do
+  use Koans
+
+  @intro "Streams"
+
+  koan "Streams are lazier than Enum, they only return a struct" do
+    assert Stream.map([1, 2, 3], &(&1 + 1)).__struct__ == ___
```
iamvery: As a learning, I wouldn't know what the form of the structure is here. It might also be nice to explain _why_ a structure is returned. Something like "Unlike Enum, a Stream does not execute immediately. It instead returns a struct the represents the enumeration to be performed later". I'm not a huge fan of that wording, but does is my point clear?

It might also be nice to initially have an example like "Streams iterate over collections and Enums" followed by something like "But they're lazy".

lib/koans/19_streams.ex
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  koan "You have to ask it explicitly to do the operation" do
```
iamvery: I believe the term often used for these operations are "terminal operations". So you might word this like "Streams are executed by terminal operations"

lib/koans/19_streams.ex
```diff
@@ -0,0 +1,27 @@
...
+    updated_list =
+      [1, 2, 3]
+      |> Stream.map(&(&1 + 1))
```
iamvery: We're missing something important in these examples, an illustration that streams avoid iterating over a collection multiple times. That might be hard to illustrate, do you have any ideas? At the very least, we should probably have an example that shows piping multiple Stream calls together to make the point that each one does not represent an iteration through the collection, but rather a "stacking" of the streamed behavior in the pipeline.

lib/koans/19_streams.ex
felipesere: Hmm... I wonder how I can demonstrate laziness? 🤔

lib/koans/19_streams.ex
felipesere: One way I can think of this is to play with side effects by having the individual 'transformations' that get stacked send a value to a pid and then inspecting the order in which they were received?

fabientownsend: The point of this example was more to show how to execute an operation on a Stream. In this very case, even if the use of a Stream isn't ideal it's still a possibility.

That's why I preferred to create another example that demonstrates that you can compute a selected part of the Stream.

lib/koans/19_streams.ex
fabientownsend: 👍

lib/koans/19_streams.ex
trptcolin: @felipesere Throw some IO in the mapping function?

PR #183, opened 2017-03-31
Use Elixir 1.3 as the new minimum bar

felipesere:
@iamvery or @ukutaht merge this if you agree 😄


iamvery:
This makes sense to me. We suspect folks are more likely to use a more recent version of Elixir with the koans. If they do try to use them with a version < 1.3, what will it do?

PR #186, opened 2017-04-05
Add string interpolation koan. Fix #184

### LINKED ISSUE #184: String interpolation is first presented in the sigil lesson but not in strings lesson
I think that string interpolation can be instroduced in `02_strings.ex` instead of showing it for the first time in `03_sigils.ex`. I will be happy to propose a PR.

Nagasaki45:
I'm really not sure about the wording. I'm not a native English speaker, and believe that someone will be able to find a better text for this that will also fit with the rest of the nice koans.

[Closes #184]

### DIFF
--- a/lib/koans/02_strings.ex
+++ b/lib/koans/02_strings.ex
@@ -7,6 +7,10 @@ defmodule Strings do
     assert "hello" == ___
   end
 
+  koan "Values may be inserted into strings by interpolation" do
+    assert "1 + 1 = #{1 + 1}" == ___
+  end
+
   koan "They can be put together" do
     assert "hello world" == ___ <> "world"
   end
--- a/test/koans/strings_koan_test.exs
+++ b/test/koans/strings_koan_test.exs
@@ -5,6 +5,7 @@ defmodule StringTests do
   test "Strings" do
     answers = [
       "hello",
+      "1 + 1 = 2",
       "hello ",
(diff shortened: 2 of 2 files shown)


lib/koans/02_strings.ex
```diff
@@ -38,4 +38,8 @@ defmodule Strings do
   koan "Strings can be louder when necessary" do
     assert String.upcase("listen") == ___
   end
+
+  koan "They also support interpolation" do
```
iamvery: It might be nice to introduce this a little earlier. Perhaps just under "They can be put together". Thoughts? Also w.r.t. wording, it sounds pretty good, but we might emphasize what interpolation means. Maybe something like "Values may be inserted into strings by interpolation". Is that any clearer?

Nagasaki45: Yes, your suggested wording is much better!
I also though about putting it earlier, but after "They can be put together" we have "Or pulled apart when needed". It's good to keep them one after the other.
What about putting the new koan before "They can be put together"?

iamvery: I don't have a strong opinion. Maybe @felipesere?

felipesere: Putting them ahead sounds goods to me 👍

Nagasaki45:
I updated the PR with your suggestions.

PR #187, opened 2017-04-09
Remove 'I think I need my number as a char' koan. Fix #176

### LINKED ISSUE #176: About koan "I think I need my number as a char"
The concept of char lists was not presented before this koan. Strings, however, were already presented.
I think that this specific koan imposible to grasp without start googling for char lists and the special way they are defined in elixir (compared to python, my main language, for example).
Maybe it is too early to present it or there is a need to present some related concepts before. What do you think?

Nagasaki45:
[Closes #176]

### DIFF
--- a/lib/koans/04_numbers.ex
+++ b/lib/koans/04_numbers.ex
@@ -23,12 +23,6 @@ defmodule Numbers do
     assert number == ___
   end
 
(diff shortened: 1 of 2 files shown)


iamvery:
I think it makes sense to remove this until (and if) we introduce char lists explicitly.

PR #189, opened 2017-04-22
Simplify atoms koans. Fix #185

### LINKED ISSUE #185: Hard to understand the purpose of atoms
I think that with the current atoms koan (`05_atoms.ex`) it is hard to understand the purpose of them. OK, they're "like strings", but what shoud we do with them. The fact that `true`, `false`, and modules (which were'nt instroduced yet, BTW) are atoms is even more confusing.

First, if we want to compare atoms to something it might be better to compare them to enums (like C enums).
Second, I think that instead of being correct, as with the last two atom koans (the ones about modules), we can just state that atoms are only equal to themselves, give a naive example, and move on.

Nagasaki45:


lib/koans/05_atoms.ex
```diff
@@ -3,29 +3,15 @@ defmodule Atoms do
...
 
-  koan "Strings can be converted to atoms, and vice versa" do
-    assert String.to_atom("atomized") == ___
-    assert Atom.to_string(:stringified) == ___
-  end
-
   koan "It is surprising to find out that booleans are atoms" do
     assert is_atom(true) == ___
-    assert is_atom(false) == ___
+    assert is_boolean(false) == ___
```
Nagasaki45: I choose to add this, as it might help to make the point even clearer.

ukutaht:
Thank you for the contribution 👍

Issue #190, opened 2017-04-24
Koans for Sigil references lists before the concept of a "list" is introduced


tjarratt:
This might be a bit pedantic, and come across as whiny, so I'd like to open this issue with a hearty "THANK YOU" for creating and maintaining this project. I discovered this project last night after a round of doing some Kotlin koans (and complaining how bad the user experience was with `gradle` compared to the ruby koans with `rake`). I was really blown away by how delightful it is to run `mix meditate` and stay in a happy state of flow while learning about elixir. 

While I was doing exercise 3, I found that there was [some knowledge needed to complete sigils that hadn't been introduced yet, namely lists](https://github.com/elixirkoans/elixir-koans/blob/master/lib/koans/03_sigils.ex#L27).

I was able to get around this by referring to exercise 2, where strings were split up, but it felt a little awkward and strange compared to the cadence of other koans within the previous two exercises.

Is it desirable to cause the user to slow down drastically here in this manner?

Nagasaki45:
Good point! I actually think that it will be better to move the whole sigils lesson after all of the data types lessons. I see no reason for strings to be explained before sigils, and lists to be exaplained after it. What do you say about moving it after the structs lesson?

tjarratt:
So the proposed order up to structs would be this: (ignoring later lessons for now)

* 01_equalities.ex
* 02_strings.ex
* 03_numbers.ex
* 04_atoms.ex
* 05_tuples.ex
* 06_lists.ex
* 07_keyword_lists.ex
* 08_maps.ex
* 09_map_sets.ex
* 10_structs.ex
* 11_sigils.ex

This seems pretty good, in that it takes a useful language feature (sigils) and delays teaching it until after a lot of other core concepts. By this point, the user certainly has a better grasp of other fundamental language features (strings, integers, functions) that will be similar to other features in languages they may already know.

Another possible solution would be to take the list-specific features from `sigil.ex` and move them into the list lesson. 

I'm happy to issue a PR in either case, just let me know what you think would be best and I'll happily contribute a fix. 😄

Nagasaki45:
I'm not one of the project maintainers, so it's not for me to decide :-)
But yes, this is what I had in mind.

tjarratt:
lol that feeling when you assume the first person to respond to your github issue is the maintainer 

"whoops"
![whoops](https://i.imgur.com/m1IctIR.png)

felipesere:
[...]

PR #192, opened 2017-04-25
Move sigils to later in the lessons

tjarratt:
While doing this, I also discovered that there is also a reference to lists in numbers, but that might a bit easier for someone to grasp, given the hint that the koan gives, and the output they see when they run it.

Also added a small hint in the strings exercise to help the user remember that lists are a thing that exist, even if they haven't been taught about them yet.


tjarratt:
This is for #190.

PR #193, opened 2017-04-27
Convert to application

marosluuce:
This extracts a number of the stateful modules into an application supervision tree. The koan running itself can stand as an example of how to architect an application using workers in a supervision tree. There are some more ideas on how to take this further that I've discussed with @felipesere.


felipesere:
@marosluuce on `master` code was added to make the koans pick up files from the filesystem rather than have them hard-coded. For this branch, would you suggest having another otp-application that deals with loading them? Or would you keep that in the runner.ex?

marosluuce:
I'm not sure whether it warrants its own watching loader. Could definitely be added in the future. Right now, it doesn't seem slow to re-check the filesystem.

PR #199, opened 2017-07-11
Fetching values of repeated key in keywordlist

codefupanda:
Koan 7.3 is not clear. Updated to make it generic and use List module instead of pattern matching (which is not yet introduced to the learner)

### DIFF
--- a/lib/koans/07_keyword_lists.ex
+++ b/lib/koans/07_keyword_lists.ex
@@ -15,11 +15,11 @@ defmodule KeywordLists do
     assert kw_list[:foo] == ___
   end
 
-  koan "You could access a second key by removing the first" do
-    kw_list = [foo: "bar", foo: "baz"]
-    [_|kw_list] = kw_list
+  koan "You could access the values of repeating key" do
+    kw_list = [foo: "bar", foo1: "bar1", foo: "baz"]
+    values = Keyword.get_values(kw_list, :foo)
 
-    assert kw_list[:foo] == ___
+    assert List.last(values) == ___
   end
 
   koan "Keyword lists are just special syntax for lists of two-element tuples" do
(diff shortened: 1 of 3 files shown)


felipesere:
Hi codefupanda! 
Thank you for this PR ❤️💛💚💙💜!

I feel like your suggestion looses the intent of the koan, which is to use pattern matching to access data deep in the list.
Is there a way we could improve the wording or example to keep the pattern match, but make the intent more obvious?
Maybe showing both ways (pattern and Keyword module) to access the second field of a keyword list?

codefupanda:
My understanding is learner is not yet introduced to the pattern matching. Pattern matching might be more useful when we know the number of occurrences of the key in advance, which is not an usual case with a keyword list. A generic solution is what Keyword.get_values already implements.

```elixir
koan "You can pattern match, if you know the order" do
    kw_list = [foo: "bar", foo1: "bar1", foo: "baz"]
    [foo: value1, foo1: "bar1", foo: value2] = kw_list
    assert value2 == ____
end
````

Curious: why no patter matching koan(s) in tuple or list chapters?
Opinion: Good to keep all the pattern matching koans in 12_pattern_matching.ex

Let me know your opinion.

Nagasaki45:
I tend to agree with @codefupanda. For me, the most obvious use cases for pattern matching are tuples and maps. So I think it will be better to either: show some pattern matching earlier, in the tuples lesson, and add some pattern matching examples to the maps lesson (it might make the pattern matching lesson unnecessary), or, as @codefupanda rightfully suggests, keep the pattern matching examples for later.

Just my 2p...

felipesere:
@Nagasaki45 Pattern-matching on keyword-lists is a great way to add named arguments to functions.

I am happy with merging this PR! Thank you both @codefupanda and @Nagasaki45 ❤️ 💚 💙

PR #200, opened 2017-07-11
Get lessons order from filenames. Fix #196

### LINKED ISSUE #196: Make sure the runner and filenames agree on koans order
I open this issue in response to https://github.com/elixirkoans/elixir-koans/pull/194. It's a good catch of a a trivial bug we missed in https://github.com/elixirkoans/elixir-koans/pull/192, when we changed the order of the koans (the filenames) without changing the order of the modules in `runner.ex`.

As I see it, to make sure we don't mess this up again we have two options:
1. Have a test that validate the filenames against the order of the modules in `runner.ex`.
2. Make `runner.ex` build the order from the order of the filenames.

If you have other suggestions do tell.
I will be happy to (try to) implement a solution with either approach.

Nagasaki45:
[Closes #196]

### DIFF
--- a/lib/runner.ex
+++ b/lib/runner.ex
@@ -1,32 +1,31 @@
 defmodule Runner do
-  @modules [
-    Equalities,
-    Strings,
-    Numbers,
-    Atoms,
-    Tuples,
-    Lists,
-    KeywordLists,
-    Maps,
-    MapSets,
-    Structs,
-    Sigils,
-    PatternMatching,
-    Functions,
-    Enums,
-    Processes,
-    Tasks,
-    Agents,
-    GenServers,
-    Protocols,
-  ]
 
-  def koan?(koan), do: Enum.member?(@modules, koan)
+  def koan?(koan) do
+    Keyword.has_key?(koan.__info__(:exports), :all_koans)
+  end
 
-  def modules, do: @modules
+  def modules do
+    {:ok, modules} = :application.get_key(:elixir_koans, :modules)
+
+    modules
+    |> Stream.map(&(&1.module_info |> get_in([:compile, :source])))
+    |> Stream.map(&to_string/1)  # Paths are charlists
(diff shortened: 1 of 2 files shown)


Nagasaki45:
Waiting to hear your opinion and add changes if necessary.

felipesere:
Thank you for your contribution! ❤️💛💚💙💜
I'll take a closer look tonight when I'm home.

lib/runner.ex
```diff
@@ -1,31 +1,31 @@
...
+    |> Stream.map(&(&1.module_info |> get_in([:compile, :source])))
+    |> Stream.map(&to_string/1)  # Paths are charlists
+    |> Stream.zip(modules)
+    |> Stream.filter(fn {_path, mod} -> koan?(mod) end)
+    |> Stream.map(fn {path, mod} -> {path_to_number(path), mod} end)
+    |> Enum.sort_by(fn {number, _mod} -> number end)
+    |> Enum.map(fn {_number, mod} -> mod end)
+  end
+
+  @koan_path_pattern ~r/lib\/koans\/(\d+)_\w+.ex$/
```
felipesere: Will this fail on Windows? I don't have a system available to try, but I feel that might cause issues...

Nagasaki45: Oh... I didn't thought about that. I can try it on a windows machine on Monday. If it does fail we can just ignore the entire path and match against `(\d+)_\w+.ex$`. With the `koan?` check this should be sufficient.

Nagasaki45: Checked it and it works on windows as well. `MyModule.module_info[:compile][:source]` is a unix like path, even on windows. I guess that paths in general are always unix like anyway, as `File.read("lib/runner.ex")` also works.

felipesere:
@Nagasaki45 Do you mind updating with the latest from master?

Nagasaki45:
PR updated.

PR #201, opened 2017-07-13
Added GenServer Koans

EricAnderson-AAI:


lib/koans/19_genservers.ex
```diff
@@ -0,0 +1,141 @@
...
+    end
+  end
+
+  koan "Servers that are created and initialized successfully returns a tuple that holds the PID of the server" do
+    {:ok, pid} = GenServer.start_link(BicycleLock, nil)
+    assert is_pid(pid) == ___
+  end
+
+  koan "When starting a GenServer you can set it's initial state" do
+    {:ok, pid} = GenServer.start_link(BicycleLock, "Hey Arnold!")
```
iamvery: Might be nice for the initial value to _look like_ a password. I'm thinking something like `s3kr3t!`.

lib/koans/19_genservers.ex
```diff
@@ -0,0 +1,141 @@
...
+  koan "A server can support multiple actions by implementing multiple handle_call functions" do
+    {:ok, pid} = GenServer.start_link(BicycleLock, nil)
+    assert GenServer.call(pid, :get_bike_brand) == ___
+    assert GenServer.call(pid, :get_bike_name) == ___
+  end
+
+  koan "A handler can return multiple values" do
+    {:ok, pid} = GenServer.start_link(BicycleLock, nil)
+    {:ok, multi_values} = GenServer.call(pid, :get_multi)
+    assert multi_values == ___
```
iamvery: To me this is less about "multiple values" and more about "any type of data, including lists". Is that fair?

lib/koans/19_genservers.ex
```diff
@@ -0,0 +1,141 @@
...
+  end
+
+  koan "The handle_cast callback handles asynchronous messages" do
+    {:ok, pid} = GenServer.start_link(BicycleLock, nil)
+    GenServer.cast(pid, {:reset, 1234, "Hello"})
```
iamvery: I had a hard time figuring out where `1234` came from at first. Maybe initialize the server with a value so it's clearly being repeated in the cast?

iamvery: Also would the message `:change` or even `:change_password` make the intent clearer?

lib/koans/19_genservers.ex
```diff
@@ -0,0 +1,141 @@
...
+    assert GenServer.call(pid, :get_password) == ___
+  end
+
+  koan "Handlers can also return error responses" do
+    {:ok, pid} = GenServer.start_link(BicycleLock, nil)
+    assert GenServer.call(pid, {:unlock, 2017}) == ___
+  end
+
+  koan "Referencing processes by their PID gets old pretty quickly, so let's name them" do
+    {:ok, pid} = GenServer.start_link(BicycleLock, nil, name: :bike_lock)
```
iamvery: No reason to save the `pid` in this case, right? Might be clearer to just `{:ok, _} = ` to emphasize that "we don't care what the pid is"

lib/koans/19_genservers.ex
```diff
@@ -0,0 +1,141 @@
...
+    assert GenServer.call(:bike_lock, :name_check) == ___
+  end
+
+  koan "Our server works but it's pretty ugly to use; so lets use a cleaner interface" do
```
iamvery: It might be clearer to say something like "It's tiring to repeat GenServer.start_link. A common pattern is adding a function to handle this."

iamvery: I'm also wondering if `start_link` is a leaky abstraction? Is there another name that would be more intention revealing or is this a common enough pattern to leak through the GenServer details?

lib/koans/19_genservers.ex
```diff
@@ -0,0 +1,141 @@
+defmodule GenServers do
+  use Koans
+
+  @intro "GenServers"
+
+  defmodule BicycleLock do
+    use GenServer
+
+    @init_password 1234
```
iamvery: I'm thinking having a default password is a little distracting. Why not just always require them to specify a password so that usage is explicit in the examples?

EricAnderson-AAI:
Thanks for looking over my pull request!

I have updated my code with your suggestions. 

As for the where the GenServer koans fit; in my opinion I think they should appear after the 15_processes koans. Thoughts?

lib/koans/19_genservers.ex
```diff
@@ -0,0 +1,137 @@
...
+
+  @intro "GenServers"
+
+  defmodule BicycleLock do
+    use GenServer
+
+    #####
+    # External API
+
+    def start_server(init_password) do
```
Nagasaki45: I think it's a convension to keep the name of this function `start_link`. It makes your gen servers more "supervisors" friendly :-)

lib/koans/19_genservers.ex
```diff
@@ -0,0 +1,137 @@
...
+  end
+
+  koan "When starting a GenServer you can set it's initial state" do
+    {:ok, pid} = GenServer.start_link(BicycleLock, "Hey Arnold!")
+    assert GenServer.call(pid, :get_password) == ___
+  end
+
+  koan "The handle_call callback is synchronous so it will block until a reply is received" do
+    {:ok, pid} = GenServer.start_link(BicycleLock, "3kr3t!")
+    assert GenServer.call(pid, :get_password) == ___
```
Nagasaki45: The fact that `call`s are synchronous are not really visible in this example. For us, it is clear that it must be synchronous if we want a return value, but a newcommer might miss this and end up confused.

lib/koans/19_genservers.ex
EricAnderson-AAI: I completely agree, I didn't even consider this. Will change this back.

lib/koans/19_genservers.ex
EricAnderson-AAI: Ha! good catch. This fell through the cracks. Will change this.

EricAnderson-AAI:
Hello!

I have made modifications to the code based off of your comments; also I played around with the sync/async koans but I have yet to come up with a simple example. Any suggestions?

Looking forwards to your feedback.

lib/koans/19_genservers.ex
```diff
@@ -0,0 +1,153 @@
...
+
+  koan "The handle_cast callback handles asynchronous messages" do
+    {:ok, pid} = GenServer.start_link(Laptop, "3kr3t!")
+    GenServer.cast(pid, {:change_password, "3kr3t!", "Hello"})
+    assert GenServer.call(pid, :get_password) == ___
+  end
+
+  koan "Handlers can also return error responses" do
+    {:ok, pid} = GenServer.start_link(Laptop, "3kr3t!")
+    assert GenServer.call(pid, {:unlock, 2017}) == ___
```
iamvery: Is it confusing that the attempted password (`2017`) looks like a year or something? Maybe it should be the same type and resemble a password?

EricAnderson-AAI: Agreed. Will change

EricAnderson-AAI:
@Nagasaki45 @ukutaht @felipesere @iamvery where do you guys think this fits in the koans list? In my opinion I think GenServers should come after the Process lesson. 

Also, are there any other changes I need to implement in order to get this PR merged?

Thanks for the feedback guys!

felipesere:
This is brilliant work! 
How about slotting it in after Tasks and Agents? They are more simplified abstractions over processes/GenServer to begin with... I'll happily merge it after this 👍

EricAnderson-AAI:
@felipesere I have made the change. Thanks again everyone for the feedback!

felipesere:
💙 ❤️ 💚 💛

PR #204, opened 2017-09-01
Koans for Docker users

denheck:
Really like the work you guys are doing here! I don't have elixir installed locally so I use docker for my projects. Thought I would contribute the Dockerfile and commands I have been using. From within the elixir-koans directory:
```
docker build -t elixir-koans . && docker run --rm -v `pwd`:/elixir-koans -ti elixir-koans
```

I would be happy to update the README with these commands. Thanks!

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+lib/koans
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 lib/koans
+_build
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM elixir:1.5
+RUN apt-get update && apt-get install -y inotify-tools
+WORKDIR /elixir-koans
+ADD . /elixir-koans/
+RUN mix local.hex --force 
+RUN mix deps.get 
+CMD mix meditate
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,6 @@ FROM elixir:1.5
 RUN apt-get update && apt-get install -y inotify-tools
 WORKDIR /elixir-koans
 ADD . /elixir-koans/
-RUN mix local.hex --force 
-RUN mix deps.get 
+RUN mix local.hex --force
+RUN mix deps.get
 CMD mix meditate
--- a/docker.sh
+++ b/docker.sh
@@ -1,0 +1,2 @@
+#!/bin/sh
+docker build -t elixir-koans . && docker run --rm -v `pwd`:/elixir-koans -ti elixir-koans


Nagasaki45:
Nice idea @denheck!
I'm not one of the maintainers, but saw your PR and have some thoughts and comments.

First, the fact that the user modifies the source code on the fly (by solving the koans) cause the build of the docker image to run every time. Everything is cached until the call to `ADD . /elixir-koans/`, but later steps run again and again with no reason and slow down the "startup time". I see two possible solutions:

- Instruct the user to build once and run multiple times.
- Add the `lib/koans` directory to `.dockerignore`, so the build won't contain these files and therefore won't change between builds.

Will be happy to hear about other options.

Second question. Is the `` `pwd` `` trick the best solution for the volume host source? I genuinely have no idea, but it looks like a hacky solution for a trivial docker use case.

Third, a simpler way to call this command will be nice. Maybe a script or a Makefile. As a linux user I have no idea what a crossplatform solution would look like.

Lastly, note that there are some trailing white spaces in the Dockerfile.

Will be happy to discuss this further. Let's see what the maintainers think :-)

denheck:
Hey @Nagasaki45, thanks for your feedback! 

1. You are correct about the `ADD . /elixir-koans/` layer not being cached on repeat builds when `lib/koans` changes. I have been using the build once run multiple times strategy. I think it's a good idea to include a `.dockerignore` as well so I will update this PR.

2. I am not sure of a better way to do this but am open to suggestions. I borrowed the `pwd` technique for mounting volumes from the [Docker docs](https://docs.docker.com/engine/reference/commandline/run/#mount-volume--v-read-only). 

3. I am not sure how to simplify this command without requiring additional software for the end user to install. A possible solution is Docker Compose which ships with Docker for Windows and Docker for Mac.

4. I will clean up the whitespace.

I appreciate your comments and suggestions. Thanks again!

Nagasaki45:
Thanks for the update!

1. I pulled your changes and tried them and it still rebuild the image from the same place. Maybe the code is compiled on the fly and new files are written to the `_build` directory. I'm not sure...

1. So maybe that's the proper way. It just seemed wierd to me.

1. I thought about a Makefile or a script. But these are platform specific, aren't they? On the worse scenario we can write a mix task for that. Something like `mix docker meditate` :wink:.

1. Perfect!

Thanks again

denheck:
Nice catch on adding the `_build` folder to `.dockerignore`! Thanks for finding that issue! I updated the PR.

I don't think Make or Bash come pre-installed on Windows. Unfortunately we can't use mix to solve it either without installing Elixir on the host machine.

Let me know if you find any other issues. Thanks again for your feedback!

Nagasaki45:
> Unfortunately we can't use mix to solve it either without installing Elixir on the host machine.

When you are so concentrated on the details that you forget the porpuse :grimacing:

felipesere:
This looks pretty nifty! Thx!
I'd suggest a very basic `shell` script to begin with. Windows for Docker isn't _that_ popular at the moment. When it becomes an issue, we can add a `batch` file.

denheck:
Thanks for the feedback @felipesere! I added the docker commands mentioned above to a shell script.

DaneWeber:
Any outstanding actions before this can be merged?

PR #205, opened 2017-09-04
Create Foo

justanirudh:


iamvery:
It appears that this PR was submitted in mistake. No worries @justanirudh!

felipesere:
Does this mean we might get a different PR, with some (even small!) improvement on the koans? 😄

justanirudh:
Ah, my bad guys.
Sorry about that.

PR #207, opened 2017-09-19
Replaced spawn_link call with a simple spawn

danielemegna:
spawn_link call seems not needed here.
I tried it (Elixir 1.5.1) and the test seems to fails and succeeds exactly in the same way.

Is there another reason that I'm ignoring?


iamvery:
I suspect the purpose is two part:
1. Ensure the process is not orphaned and lingering around in the background for the remainder of your run.
2. Ensure the parent process shuts down if there happens to be a problem in the spawned process.

Does this sound reasonable? Thanks for your question!

PR #208, opened 2017-10-27
Use more succinct syntax for mapping with String.capitalize

mkwiatkowski:
It's more idiomatic and that syntax was introduced two koans earlier, so it should be OK to use it.


felipesere:
Thank you ❤️ 💚 💙

PR #209, opened 2017-11-07
Fix __info__ for the upcoming Elixir 1.6

AMar4enko:


ukutaht:
Thanks for the PR! 

Could you provide a reference as to what change in the language this accommodates? Is it backwards compatible with 1.3 - 1.5?

iamvery:
This also reminds me that we should probably go ahead and add 1.5 (and 1.6 when available) to the .travis.yml build matrix.

AMar4enko:
@ukutaht 
https://github.com/elixir-lang/elixir/commit/5ed491a7ea8962310a5a5763bf358d42db58a421#diff-bfef8c8eee73046f4daf3abc1822e321

Previous doc version statement one can pass any atom besides defined by Elixir itself, supported by erlang module_info/1 is now gone. And it does seem as a breaking change.

ukutaht:
Fixes #211 

Thank you for the contribution, @AMar4enko ❤️  !

PR #217, opened 2018-03-01
introduce MapSet.size/1 earlier

Cohen-Carlisle:
specifically, before it is used to show sets do not have duplicates

Edit: Instead, remove the later `MapSet.size/1` koan as the function is sufficiently introduced by `koan "However, I do not allow duplication"`.

### DIFF
--- a/lib/koans/09_map_sets.ex
+++ b/lib/koans/09_map_sets.ex
@@ -9,6 +9,10 @@ defmodule MapSets do
     assert Enum.fetch(@set, 0) == {:ok, ___}
   end
 
+  koan "How large is my map set?" do
+    assert MapSet.size(@set) == ___
+  end
+
   koan "However, I do not allow duplication" do
     new_set = MapSet.new([1, 1, 2, 3, 3, 3])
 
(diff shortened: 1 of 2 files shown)


lib/koans/09_map_sets.ex
```diff
@@ -9,6 +9,10 @@ defmodule MapSets do
     assert Enum.fetch(@set, 0) == {:ok, ___}
   end
 
+  koan "How large is my map set?" do
```
iamvery: I don't like that this koan moved here interrupts the conversation of the others it falls between.

"I am very similar to a list" -> "However, I do not allow duplication"

Regardless, you are definitely onto something as "However, I do not allow duplication" uses MapSet.size/1 before it's introduced. What do you think about just removing the explicit "How large is my map set?" altogether and let that be something to glean from "However, I do not allow duplication"?

Cohen-Carlisle: Sounds good to me!

felipesere:
Right now the diff looks like it is just removing the koan? Is that correct?

Cohen-Carlisle:
Yes, I'll update the description. @iamvery suggested (in the outdated review comment) removing the later `MapSet.size/1` koan since that function is actually introduced by `koan "However, I do not allow duplication"` and I agreed with this approach.

PR #218, opened 2018-03-02
more number koans

### LINKED ISSUE #216: Should `===` be used for differentiating floats and ints?
Here's the example that made me think this might be a good idea in some cases:
https://github.com/elixirkoans/elixir-koans/blob/18ce1752d45c9069905e31f3e67ad79259edc1d2/lib/koans/03_numbers.ex#L52-L54
`assert Float.ceil(34.25) == 35` would pass this test, even though `Float.ceil/1` produces a float.

Cohen-Carlisle:
Closes #216

### DIFF
--- a/lib/koans/03_numbers.ex
+++ b/lib/koans/03_numbers.ex
@@ -4,6 +4,38 @@ defmodule Numbers do
 
   @intro "Why is the number six so scared? Because seven eight nine!\nWe should get to know numbers a bit more!"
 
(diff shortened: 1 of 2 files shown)


Nagasaki45:
Nice! I learned a few things here :-)

Cohen-Carlisle:
I realized I forgot about `Kernel.rem/2`, so added a koan for that as well.

felipesere:
Thank you! ❤️ 💚 💙

PR #219, opened 2018-03-27
Add another `assert` to Agent "maintain state" koan

jordan-brough:
To clarify/emphasize that this is a persistent state.

### DIFF
--- a/lib/koans/17_agents.ex
+++ b/lib/koans/17_agents.ex
@@ -6,6 +6,7 @@ defmodule Agents do
   koan "Agents maintain state, so you can ask them about it" do
     {:ok, pid} = Agent.start_link(fn -> "Hi there" end)
     assert Agent.get(pid, &(&1)) == ___
+    assert Agent.get(pid, &(&1)) == ___
   end
 
   koan "Agents may also be named so that you don't have to keep the pid around" do


jordan-brough:
When I first saw the code it wasn't 100% clear to me whether the result was a one-time thing or not. I wrote this extra test to confirm to myself that the state would remain persistently, and thought perhaps it would be a helpful thing to have by default.

iamvery:
Hmm ya, I see what you mean. I feel like even this example might lead the learner to believe that the function is executed each time rather than being a producer of the state. Do you think a later example that shows how the state can be updated serves to solidify that better than a repeated assertion?

Thoughts @felipesere ?

jordan-brough:
For me this koan was confusing because the text _said_ "maintain state, so you can ask them about it" but the example only queried a single time so the persistence wasn't clear.  Especially coming fresh from the previous `Processes` and `Tasks` modules (where messages and responses are one-time events) this aspect was confusing to me initially.  It would be nice to have something that clears that up imo.  The later examples do but I was confused about that initially.

felipesere:
Thank you @jordan-brough for opening this PR 👍 

Should we maybe connect this Koan somehow with one where we explicitly mutate the state?
Its kinda hard to prove that the state won't change. Looking at it now (having done a bunch of Elixir since the Koans) it feels natural that nothing changes unless instructed to do so. 
Maybe your connection to processes and messages also gives us a clue? Maybe we can make the difference more explicit?

@iamvery I also feel we could extract the function in `start_link` itself into something like
`one_off_producer = fn -> state end` and lift `state` into its own variable?

iamvery:
@felipesere i like that! let the code communicate that intent 💪

jordan-brough:
Sounds good!  I'll take a stab at refactoring along those lines. Thanks.

Issue #53, opened 2016-04-04
Koans for :with keyword


felipesere:
See https://github.com/elixir-lang/gen_router/blob/2001c015b560d82f8077f323de5ed0103f5e0566/lib/dynamic_supervisor.ex#L398-L404 for an example.

s-oram:
Is anyone working on this? I could try putting something together for it. @felipesere

felipesere:
Go for it!

s-oram:
I've just sent a pull request with 6 koans to meet item one on the list above (the `for` keyword). Please let me know if you see anything that can be improved. 

I started working a while back so haven't been super motivated to do extra programming on the weekends. Normally I'd rather get outside and play some sport. (It just happens to be raining this weekend.) If you're happy with a slow pace, I'll continue working on koans for the remaining items when time and energy permits. :) 

Thanks Felipersere!

PR #225, opened 2018-09-02
Add new koans for comprehensions

s-oram:
The koans added here introduce comprehensions. (the `for` keyword.)

"Koans for special forms" https://github.com/elixirkoans/elixir-koans/issues/53

### DIFF
--- a/lib/koans/20_comprehensions.ex
+++ b/lib/koans/20_comprehensions.ex
@@ -1,12 +1,9 @@
 defmodule Comprehensions do
   use Koans
 
-  @intro "Comprehensions"
+  @intro "A comprehension is made of three parts: generators, filters, and collectibles. We will look at how these interact with eachother"
 
-  # A comprehension is made of three parts: generators, filters and collectables.
-
-  koan "Generators provide the values to be used in a comprehension" do
-    # In the expression below, `n <- [1, 2, 3, 4]` is the generator.
+  koan "The generator, `n <- [1, 2, 3, 4]`, is providing the values for our comprehension" do
     assert (for n <- [1, 2, 3, 4], do: n * n) == ___
   end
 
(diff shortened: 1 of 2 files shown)


s-oram:
Please let me know if there is anything that can be improved. Thanks!

lib/koans/20_comprehensions.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Comprehensions do
+  use Koans
+
+  @intro "Comprehensions"
+
+  # A comprehension is made of three parts: generators, filters and collectables.
```
felipesere: This would be a lovely `@intro`. Maybe even expanded to something like
```
@intro "A comprehension is made of three parts: generators, filters, and collectibles. We will look at how these interact with eachother".
```
I know other koans don't have a fancy `@intro`... its because I put the plumbing in place but never circled back to make them fun and read like a story. These koan could be refreshing comeback to that idea :smile:

lib/koans/20_comprehensions.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Comprehensions do
+  use Koans
+
+  @intro "Comprehensions"
+
+  # A comprehension is made of three parts: generators, filters and collectables.
+
+  koan "Generators provide the values to be used in a comprehension" do
+    # In the expression below, `n <- [1, 2, 3, 4]` is the generator.
```
felipesere: Similar trick as above, can we fold the key idea of the comment up into the description?

felipesere:
The "request changes" button in Github is much harsher than what I had in mind 😢 
I really like what you have put together!!! ❤️ 💙 💚 💛 🖤

felipesere:
Thank youuuuu 👍

s-oram:
> The "request changes" button in Github is much harsher than what I had in mind

No worries Felipesere. I appreciate the guidance! :) ❤️

s-oram:
PS: There were a couple of extra changes in there too!

PR #227, opened 2018-12-03
Fixes Watcher/Autorunner for some editors (ex: TextMate)

RobertLowe:
Fixes Watcher/Autorunner for some editors (ex: TextMate)

---

Woo, creating PRs before even meditating


ukutaht:
Agreed. I trust this works without downloading TextMate to verify :)

Thank you for the contribution @RobertLowe

PR #234, opened 2019-01-07
Enhance dockerfile

krlsdu:
When you create the image the command "mix deps.get " include the dependencies in a separate layer, so, the second command, "mix meditate" launch this error:

> Unchecked dependencies for environment dev:
> _* file_system (Hex package)
>   the dependency is not available, run "mix deps.get"
> ** (Mix) Can't continue due to errors on dependencies

Because not is the same session and that layer created not is accessible
The improve resolve that problem.

The last improve is the alpine image, because of the size.
In my experiments, the size of resulted images is big different.
Not alpine base image, the size of the resulted image is 1.1GB
Using alpine base image the size of the resulted image is 90.3MB

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,7 @@ matrix:
       otp_release: 19.1
     - elixir: 1.5.1
       otp_release: 20.0
-after_success: 
-  - "curl -H \"Content-Type: application/json\" --data '{\"build\": true}' -X POST https://registry.hub.docker.com/u/krlsdu/elixir-koans/trigger/${TOKEN_DOCKER_BUILD}/"
+    - elixir: 1.6.5
+      otp_release: 20.0
+    - elixir: 1.7.3
+      otp_release: 21.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,4 @@ matrix:
     - elixir: 1.5.1
       otp_release: 20.0
 after_success: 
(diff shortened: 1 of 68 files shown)


krlsdu:
Ok @iamvery.

README.md
```diff
@@ -1,6 +1,9 @@
 # Elixir Koans
 
-[![Build Status](https://travis-ci.org/elixirkoans/elixir-koans.svg?branch=master)](https://travis-ci.org/elixirkoans/elixir-koans)
+[![Build Status](https://travis-ci.org/krlsdu/elixir-koans.svg?branch=master)](https://travis-ci.org/krlsdu/elixir-koans)
```
felipesere: This and the other badges seem like a leftover. 
We might be interested to pull some of those things over to the elixirkoans organisation though 🤔

.travis.yml
```diff
@@ -1,3 +1,4 @@
+
```
felipesere: What does this “+” do?

felipesere:
I am not 100% sure I understand the issue with running mix deps.get in a separate layer.
My expectation was that all the changes in the file system would be reflected in later layers?
It could be tha my understanding here is wrong though 😊

I’ll give this PR a try later when I get home

.travis.yml
krlsdu: My mistake fix o commit 
4adfafb

README.md
krlsdu: My mistake fix on commit 
06187ba

PR #237, opened 2019-04-01
Remove duplicative koan

### LINKED ISSUE #236: Test duplication on GenServer koans
I think it should not be the same test. Maybe i am missing something?
```elixir
koan "When starting a GenServer you can set it's initial state" do
    {:ok, pid} = GenServer.start_link(Laptop, "3kr3t!")
    assert GenServer.call(pid, :get_password) == ___
end

koan "The handle_call callback is synchronous so it will block until a reply is received" do
    {:ok, pid} = GenServer.start_link(Laptop, "3kr3t!")
    assert GenServer.call(pid, :get_password) == ___
end
```

iamvery:
This truth is self-evident from the other koans.

[Closes #236]

### DIFF
--- a/lib/koans/18_genservers.ex
+++ b/lib/koans/18_genservers.ex
@@ -98,11 +98,6 @@ defmodule GenServers do
     assert is_pid(pid) == ___
   end
 
-  koan "When starting a GenServer you can set it's initial state" do
-    {:ok, pid} = GenServer.start_link(Laptop, "3kr3t!")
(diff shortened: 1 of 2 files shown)


iamvery:
@bontequero does this fix #236 from your perspective? thanks again for the report!

bontequero:
> @bontequero does this fix #236 from your perspective? thanks again for the report!

It does. I am glad to help!

Issue #231, opened 2019-01-02
 Range.range?/1 is deprecated


pyramind10:
In `03_numbers.ex` there is a koan for checking for range
```elixir
koan "Is this a range?" do
    assert Range.range?(1..10) == ___
    assert Range.range?(0) == ___
  end
```

`Range.range?/1` is deprecated.

This is a kind of a hard to replace because of the learning path. Behind the scenes a Range is a struct. But pattern matching and structs don't get "taught" until later in the koan journey. 

Could replace the equality check with:
```elixir
assert Range.new(1, 10) == ___  # 1..10
assert Range.new(0, 0) == ___ # 0..0
```

But I'm not sure if this really helps learn "check if it's a range". It might make more sense to remove this specific koan from `03_numbers.ex` and move a modified version to pattern matching or structs.

I think moving it to structs or pattern matching could makes more sense. 

```elixir
koan "A Range is also a struct" do
    %Range{first: first, last: last} = 1..10
    assert first == ___  # 1
    assert last == __ # 10
end
```

A few other ideas to check for a range:
`match?(%Range{}, 1..10)`
`Range{} = Range.new(1, 10)`
`%Range{} = 1..10`

Or that specific koan could be removed, which I don't see as harmful to the learning path.

iamvery:
Thanks for the report! It's probably high time that some of the maintainers give a run through the Koans to think about the journey again. It's hard to intuit these types of reports without fresh experience with the flow itself. I suspect your intuition is spot on.

pyramind10:
@iamvery no problem. Let me know if you need any help.

foucist:
I ended up fixing the deprecation warning in #253 by adding a couple of helper functions in order to keep the existing koan.  I guess it hints at the Range struct without potentially getting in the way of the learning path?

PR #240, opened 2019-06-28
Update handling of --no-clear-screen switch

filalex77:
- pass `strict:` option to [`OptionParser.parse/2`](https://hexdocs.pm/elixir/OptionParser.html#parse/2)
- use [negation](https://hexdocs.pm/elixir/OptionParser.html#parse/2-negation-switches)

---
This change is [Reviewable](https://reviewable.io/reviews/elixirkoans/elixir-koans/240)


filalex77:
After looking at #231, I think that `Range.range?/1` replacement should be discussed first.

PR #241, opened 2019-07-03
Add cloning the repo as a first step to the README

clupprich:
Might just be a small thing (and a complete lack of ☕️ in the morning), but I was missing the very first instruction in the README to actually clone the repo first, before running `mix deps.get` 🤦‍♂.


felipesere:
👍 Good catch!

PR #242, opened 2019-10-12
Remove stacktrace warning

felipesere:
Using __STACKTRACE__ is not really an option due to it requiring a
try/catch block, which we'd have to possibly add to every koan.

Side-step the issue by calling into Erlang

### DIFF
--- a/lib/execute.ex
+++ b/lib/execute.ex
@@ -40,7 +40,7 @@ defmodule Execute do
 
   defp expand({:error, stacktrace, exception}, module) do
     {file, line} =
-      stacktrace
+      :erlang.get_stacktrace()
       |> Enum.drop_while(&(!in_koan?(&1, module)))
       |> List.first()
       |> extract_file_and_line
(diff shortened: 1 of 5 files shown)


iamvery (approved):

lib/execute.ex
```diff
@@ -40,7 +40,7 @@ defmodule Execute do
 
   defp expand(error, module) do
     {file, line} =
-      System.stacktrace()
+      :erlang.get_stacktrace()
```
iamvery, line 43: This will have to be rebased due to a change in #251

josevalim:
FWIW, this is not going to work on recent OTP versions. Erlang/OTP 23 always returns an empty list for stacktrace and the next version will fully remove the function.

felipesere:
@josevalim @iamvery My current thinking is that we can avoid going through the stacktrace to find file & line if we just capture them in the generated `koan` function using `__ENV__.file` and `__ENV__.line` in the first place.

I'll try a PR this evening and update this :)

felipesere:
I'm quite happy with this new approach. 
If an error happens within a Koan, I grab `__ENV__.line` and `__ENV__.file` and report it back up as part of the error. That way, there is no need to walk the stacktrace anymore!

PR #246, opened 2020-01-08
Replace deprecated Enum.into/2 for non-empty list with ++ operator

selfsimilar:


ukutaht:
Looking at the koan, I feel like it was created to teach the `for x in ..., into: ...` form. I'm thinking maybe we can keep the syntax but just collect into an empty list which is not deprecated.

I imagine the most common use-case would creating a map so:

```
collection = for x <- ["Pecan", "Pumpkin"], into: %{}, do: {x, "#{x} Pie"}
assert collection == %{"Pecan" => "Pecan Pie", "Pumpkin" => "Pumpkin Pie"}
```

What do you think about using this instead of `++`? We already teach the `++` operator in `lib/koans/06_lists.ex`.

Sorry about bringing this up after you've already created the PR. I should have reacted in the issue when you proposed the change.

selfsimilar:
TBH, I'm new to the language and don't feel confident saying any code pattern is particularly idiomatic. So I'll happily take your word that the koan was geared for teaching the `for x in ..., into: ...` for non-empty lists. I'm happy to re-roll this PR with that understanding.

PR #247, opened 2020-01-08
Remove Windows-style carriage returns from line endings

selfsimilar:
File `lib/koans/20_comprehensions.ex` is the only koan that has Windows-style `^M` carriage return line endings. This PR removes those to align with the other files.


ukutaht:
Thanks!

Issue #245, opened 2020-01-06
warning: the Collectable protocol is deprecated for non-empty lists


selfsimilar:
The final koan in comprehensions (20) throws a warning:

    warning: the Collectable protocol is deprecated for non-empty lists. The behaviour of
    things like Enum.into/2 or "for" comprehensions with an :into option is incorrect when
    collecting into non-empty lists. If you're collecting into a non-empty keyword list,
    consider using Keyword.merge/2 instead. If you're collecting into a non-empty list,
    consider concatenating the two lists with the ++ operator.
      (elixir) lib/collectable.ex:83: Collectable.List.into/1
      (elixir_koans) lib/koans/20_comprehensions.ex:29: Comprehensions."Add the
    result of a comprehension to an existing collection"/0
      (elixir_koans) lib/execute.ex:34: Execute.exec/4

This is easily remedied by using `++` instead of `into:`.

iamvery:
Thanks for the issue! Feel free to open a PR is you have a solution, otherwise someone will pick it up eventually ❤️

selfsimilar:
#246

selfsimilar:
Updated PR based on notes in #248.

PR #248, opened 2020-01-10
Use Collectable protocol with an empty map

selfsimilar:


selfsimilar:
See notes on #246 for context.

ukutaht:
Nice, thanks!

foucist:
closes #245

PR #251, opened 2020-06-04
Change usage of System.stacktrace into __STACKTRACE__ and make tests pass

### LINKED ISSUE #250: FunctionClauseError when starting newly cloned elixir-koans
This is my absolute first contact with Elixir: so I have no clue what I'm doing! I freshly installed Erlang and Elixir on Ubuntu, cloned this nice koan project, installed the mix dependencies and started to meditate. But I get an error right away. Here the list of things I have installed. Please let me know if I can provide more information!

```bash
$ elixir -v
Erlang/OTP 23 [erts-11.0.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Elixir 1.10.3 (compiled with Erlang/OTP 21)

$ mix -v
Erlang/OTP 23 [erts-11.0.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Mix 1.10.3 (compiled with Erlang/OTP 21)

$ mix deps.get
Resolving Hex dependencies...
Dependency resolution completed:
Unchanged:
  file_system 0.2.2
All dependencies are up to date

$ mix meditate
15:56:28.380 [error] Process #PID<0.459.0> raised an exception
** (FunctionClauseError) no function clause matching in Execute.extract_file_and_line/1
    (elixir_koans 0.0.1) lib/execute.ex:53: Execute.extract_file_and_line(nil)
    (elixir_koans 0.0.1) lib/execute.ex:46: Execute.expand/2
    (elixir_koans 0.0.1) lib/execute.ex:35: Execute.exec/4
```

Thank you already for your work and the koans!

kknives:
**Changes in getting a stacktrace from a failed koan**
As of Elixir 1.10, the usage of `System.stacktrace` to retrieve a stacktrace is now deprecated and is also non functional. Refer to the [docs](https://hexdocs.pm/elixir/System.html#stacktrace/0).

#242 tried substituting `System.stacktrace` with Erlang's `:erlang.get_stacktrace()` but it returns an empty list instead.

So, I substituted the calling of `System.stacktrace` in a non rescue context when trying to report errors with a call to `__STACKTRACE__` in the Koan macro itself, for normal execution and for tests, something that #242 suggested.

**Making the tests pass**
Even then, the tests would not pass. Two tests were failing.
The failure_test.exs was providing input arguments which were not correct for the output the tests asserted.
The comprehension_koans_test.exs was out of date with respect to the koan itself.

**Now the tests pass and a few koans I tried worked as well.**

[closes #250]

### DIFF
--- a/lib/execute.ex
+++ b/lib/execute.ex
@@ -38,14 +38,14 @@ defmodule Execute do
 
   defp expand(:ok, _), do: :ok
 
-  defp expand(error, module) do
+  defp expand({:error, stacktrace, exception}, module) do
     {file, line} =
-      System.stacktrace()
+      stacktrace
       |> Enum.drop_while(&(!in_koan?(&1, module)))
       |> List.first()
       |> extract_file_and_line
 
-    %{error: error, file: file, line: line}
+    %{error: exception, file: file, line: line}
   end
 
   defp in_koan?({module, _, _, _}, koan), do: module == koan
--- a/lib/koans.ex
+++ b/lib/koans.ex
@@ -22,7 +22,7 @@ defmodule Koans do
           unquote(compiled_body)
           :ok
         rescue
-          e -> e
+          e -> {:error, __STACKTRACE__, e}
         end
       end
     end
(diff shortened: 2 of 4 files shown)


marco-jantke:
Works for me now :tada: Thanks!

PR #252, opened 2020-08-18
Mismatch on koan description and assertion in 07_keyword_lists

foucist:
Koan description refers to the values of the key in a keyword list, so it should be asserting for a list of values.  

Also remove some duplication of struct defaults, %Plane already has `maker: boeing` default.


ukutaht:
LGTM 👍

PR #253, opened 2020-08-19
Fix deprecation warnings

foucist:
Shows up when running mix meditate:
`warning: not passing the :switches or :strict option to OptionParser is deprecated`

Shows up when running mix test:
`warning: Range.range?/1 is deprecated. Pattern match on first..last instead`


ukutaht:
Thanks!

foucist:
Related to #231

Issue #226, opened 2018-10-03
04_atoms.ex has unclear expectations


DaneWeber:
I'm new to Elixir and using these koans. First of all: thanks! I really appreciate them.

For the 04_atoms koans, I wasn't clear what was expected. For example, `assert true == ___` and `assert false == ___` can be most easily completed by simply repeating `true` and `false`.

I was guessing that perhaps it should be completed with `assert true == :true`.

Another thought is that these could be triple-equal, such as `assert :true === ___`?

Bottom line: while this definitely succeeding in driving me to think more about atoms, I would have appreciated more of a nudge.

If someone more experienced with Elixir can explain the intent, I would be happy to submit a PR to improve this koan.

iamvery:
@DaneWeber thank you for the feedback! Sorry it has taken so long to get a response. Sometimes it's really hard to build an example that provides enough hint to keep folks on the straight and narrow w/o giving everything away. If you have any ideas for how to help these examples explain themselves better, we'd love to see them! Maybe even as easy as a little more in the description? Or a comment?

DaneWeber:
Thanks for the response! 😄 I'm on a pretty slow cycle too.

So, I'm an elixir newbie, but my suggestion was the following:
`assert :true === ___`

I would _think_ that would point to a solution of `assert :true === true`. I'm _assuming_ that is the intent of this exercise, but I wanted to make sure I'm not missing a deeper solution.

iamvery:
Ah, I think I missed the subtlety of your suggestion. I like it! I agree that it would seem a bit easier to intuit what the lesson is trying to show you by going from atom to boolean rather than the opposite. Fwiw, I don't think the threequals is really necessary. I believe the only difference is [numeric equality](https://hexdocs.pm/elixir/1.6.0/operators.html#comparison-operators).

> The only difference between == and === is that === is stricter when it comes to comparing integers and floats

What do you think @felipesere @ukutaht ?

ukutaht:
Thank you for the suggestions @DaneWeber!

I agree with @iamvery here. Showing the atom is a good improvement to the exercise but I would keep the double-equals. In my experience the convention is to always use double-equals as the default and frankly seeing a triple-equals in an elixir codebase would be pretty confusing to me.

DaneWeber:
Interesting! I came directly from `03_numbers.ex` which uses `===` several times, and I clearly missed that it's only stricter when it comes to integers & floats. I was carrying baggage from other languages.

[...]

Issue #222, opened 2018-05-21
Keyword lists / assert_raise ArgumentError


robbyrussell:
As a first timer to Elixir, I found myself scratching my head at the last koan for KeywordLists. It's the first function and dealing with raising an argument. 

https://github.com/elixirkoans/elixir-koans/blob/08c90ce0d84120d315a8ee626ff79b5b445ececd/lib/koans/07_keyword_lists.ex#L29

Nor have I been exposed to maps so this felt like I was just trying to get it to pass without really knowing why it is. (perhaps that's part of the point of Koans to get exposed to new things without understanding them yet?)

iamvery:
This is really great feedback, @robbyrussell! Thank you! It's always hard to keep your finger on the pulse of these koans as they grow and change over time. You have to work through them over and over to really feel out the flow (which tends to just not happen in reality). So this kind of feedback from the community is really important ❤️

felipesere:
Thanks @robbyrussell for the feedback!
From a cursory look, I think we rearranged the koans and didn't pay attention to surroundings:
The first koan _also_ mentions maps, while the one you mentioned then goes  _"But unlike maps..."_.

Would it make sense to flip the koans, move them to 08_maps and then compare maps to keyword lists?

robbyrussell:
> Would it make sense to flip the koans, move them to 08_maps and then compare maps to keyword lists?

Possibly, I was just a bit thrown off by what the example was doing. (especially since it looked like I just needed to write the same thing as a few lines up but don't (yet) know why it'd trigger an ArgumentError

robbyrussell:
(side-note: was really enjoying my way through this project. Was a fun Sunday afternoon project)

s-oram:
> Possibly, I was just a bit thrown off by what the example was doing. (especially since it looked like I just needed to write the same thing as a few lines up but don't (yet) know why it'd trigger an ArgumentError

Kind of similar here. The koan subject "But unlike maps, the keys in keyword lists must be atoms" explains the salient point, but I don't understand the syntax `fn -> not_kw_list[___]`. What is `fn ->`? The koan/test passes once the assertion is rewritten to `fn -> not_kw_list[{"foo", "bar"}]` but I don't know if that is the correct answer or not. 

It's no big deal. This is my first exposure to Elixir so having to look things up to understand what's going on isn't an issue. 

[Some time later...]

[...]

PR #254, opened 2020-08-19
use atom booleans in atom koan & use keyword list in keyword list koan

foucist:
Atom koan is trying to demonstrate that booleans are actually atoms.  Writing it as `:true` and `:false` in the assertion upfront would make it more obvious, because the learner would likely be expecting to enter `true` and `false` in the blank area.  Solves #226.

Also Keyword List koan had a "non keyword list" for no reason, changing it to a keyword list keeps to the spirit of the koan and reduces confusion.  Solves #222


foucist:
(bump - updated PR description with a much better explanation of the changes)

PR #255, opened 2020-09-23
Remove Enum.fetch usage for MapSet

jechol:
`Enum.fetch(@set, 0)` usage should be discouraged as it only works for MapSet created with list size < 32.

So I think that koan should be removed.

### DIFF
--- a/lib/koans/09_map_sets.ex
+++ b/lib/koans/09_map_sets.ex
@@ -5,11 +5,7 @@ defmodule MapSets do
 
   @set MapSet.new([1, 2, 3, 4, 5])
 
-  koan "I am very similar to a list" do
-    assert Enum.fetch(@set, 0) == {:ok, ___}
-  end
-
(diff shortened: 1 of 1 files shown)


iamvery:
Good point. I see [we call out the 33-element distinction](https://github.com/elixirkoans/elixir-koans/blob/34000e5eaa0d4e162bffed2053a1deb2802478bf/lib/koans/09_map_sets.ex#L28-L30) a little further down. I think the purpose of the example is to set the stage for comparison and contrast with lists. Perhaps we should use Enum#member? instead. What do you think?

jechol:
IMHO, sets are not similar to lists, so comparison is not necessary here.

And we shouldn't teach newbie about implementation detail like magic number 32.

iamvery:
That's fair. I don't think the removal here is taking anything away. Thanks for the contribution!

PR #256, opened 2020-09-23
Change protocol examples

jechol:
I think when struct S implements protocol P, we can say S is P or S has trait of P.
For example, List implements Enumerable and we can say List is Enumerable.

In current example, Musician implements School, but we can't say Musician is School.
I think better abstraction where Musician, Dancer belongs to and Baker doesn't is Artist.

### DIFF
--- a/lib/koans/19_protocols.ex
+++ b/lib/koans/19_protocols.ex
@@ -3,49 +3,49 @@ defmodule Protocols do
 
   @intro "Want to follow the rules? Adhere to the protocol!"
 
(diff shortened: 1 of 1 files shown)


iamvery:
I don't have a strong opinion about this. @ukutaht @felipesere do you have any thoughts?

ukutaht:
I think the PR does make the example easier to understand. I think we should merge it

jechol:
@ukutaht Could you approve this?

ukutaht:
@jechol merged!

PR #258, opened 2020-12-26
Add gitpod configuration

lidorcg:


iamvery (approved):
I've not used gitpods, but I don't see any problem from this addition. Any thoughts @ukutaht @felipesere ?

felipesere:
No strong opinions other than the fact that I should read up on gitpods? Is it like Docker Compose for k8s?

PR #259, opened 2021-02-01
Add support for CodeSpaces

mattvv:
This adds a devcontainer that has elixir pre-installed so you can do the koans without having to install elixir. Selfishly useful for doing some koans while I'm playing games on my windows PC :)

To use this, you can simply clone the repo using codespaces and the terminal will automatically appear and you can get started by doing

`mix deps.get`

and then

`mix meditate`

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,18 @@
+ARG VARIANT=latest
+FROM elixir:${VARIANT}
+
+RUN apt-get update && \
+    apt-get install -y postgresql-client && \
+    apt-get install -y inotify-tools && \
+    apt-get install -y nodejs && \
+    apt-get install -y git && \
+    curl -L https://npmjs.org/install.sh | sh && \
+    mix local.hex --force && \
+    mix archive.install hex phx_new ${PHX_VERSION} --force && \
+    mix local.rebar --force
+
+ENV APP_HOME /app
+RUN mkdir $APP_HOME
+WORKDIR $APP_HOME
+
+EXPOSE 4000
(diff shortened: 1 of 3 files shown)


iamvery (approved):
I haven't used CodeSpaces myself, but I don't see any issue with this. Anyone else?

Nezteb:
@iamvery Works on my machine! (well, GitHub's machine but you get the idea 😂)

Screen Shot 2022-04-09 at 5 09 47 PM

Nezteb:
@ukutaht There is a way in the GitHub UI to do this relatively easily: https://docs.github.com/en/codespaces/developing-in-codespaces/creating-a-codespace#creating-a-codespace

Aside from that there is also the github.dev editor (which can be accessed by pressing the period key `.` on your keyboard when on a repo's page). https://docs.github.com/en/enterprise-cloud@latest/codespaces/the-githubdev-web-based-editor#opening-the-web-based-editor

As far as I can tell there is no easy way to create a Codespaces link that auto-creates one for people (similar to how Codesandbox does it).

CompuIves:
Swooping in! (I heard CodeSandbox 😂)

In the future, CodeSandbox should be able to run Elixir as well. At CodeSandbox we're currently experimenting with this (our backend is in Elixir/Phoenix), by allowing you to configure the Elixir version in [CodeSandbox Projects.](https://projects.codesandbox.io/)

Not much to add, just wanted to say that we might support Elixir in the future! 😄

Nezteb:
> (our backend is in Elixir/Phoenix)

howdidinotknowthat

iamvery:
What's the status on this?

PR #260, opened 2021-03-22
Update file_system

parkerd:
Current locked version of `file_system` doesn't work for me on a fresh clone using macOS 11.2.3.

```
MacOSX 10.7 or later required for --file-events
```

This change is the result of running `mix deps.update file_system` and allows `mix meditate` to work as expected.


ukutaht:
Thanks!

PR #263, opened 2021-07-20
be stricter about which files to watch

listx:
Things like flycheck-elixir in Emacs can create temporary files in the
same directory as the upstream files that contain the koans, with a
"flycheck_" prefix but still ending in ".ex". This change makes it so
that we skip such files.


iamvery:
Makes sense. Good addition. Thank you!

PR #271, opened 2022-02-01
Fix tests

### LINKED ISSUE #270: Fix failing tests & consider adding CI
# Problem

There are a few failing tests in the project on `master` currently. Those should be easy to address, but we may consider adding a CI to the project to avoid these going unnoticed and increase confidence in changes to the koans moving forward.

iamvery:
# To Reviewers

I'm particularly interested in your thoughts on the process isolation issue. I wasn't able to debug, but the "solution" I came up with feels a little like cheating.

# Problem

The Koan tests were failing. They probably went unnoticed, because we don't have CI set up on this project.

# Solution

1. Get the tests passing.
2. Set up CI?

[closes #270]

### DIFF
--- a/lib/koans/18_genservers.ex
+++ b/lib/koans/18_genservers.ex
@@ -134,9 +134,15 @@ defmodule GenServers do
   end
 
   koan "Our server works but it's pretty ugly to use; so lets use a cleaner interface" do
-    Laptop.start_link("EL!73")
-
+    Laptop.start("EL!73")
     assert Laptop.unlock("EL!73") == ___
+  end
+
+  koan "Let's use the remaining functions in the external API" do
+    Laptop.start("EL!73")
+
+    {_, response} = Laptop.unlock("EL!73")
+    assert response == ___
 
     Laptop.change_password("EL!73", "Elixir")
 
--- a/lib/koans/18_genservers.ex
+++ b/lib/koans/18_genservers.ex
@@ -135,15 +135,11 @@ defmodule GenServers do
 
   koan "Our server works but it's pretty ugly to use; so lets use a cleaner interface" do
     Laptop.start_link("EL!73")
-    assert Laptop.unlock("EL!73") == ___
-  end
 
-  koan "Let's use the remaining functions in the external API" do
-    Laptop.start_link("EL!73")
-    {_, response} = Laptop.unlock("EL!73")
-    assert response == ___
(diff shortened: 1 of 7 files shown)


lib/koans/18_genservers.ex
```diff
@@ -135,15 +135,11 @@ defmodule GenServers do
 
   koan "Our server works but it's pretty ugly to use; so lets use a cleaner interface" do
     Laptop.start_link("EL!73")
-    assert Laptop.unlock("EL!73") == ___
-  end
 
-  koan "Let's use the remaining functions in the external API" do
```
iamvery, line 141: ❓ The Genserver and Agent koans were timing out when running tests. It appears to have something to do with processes "leaking" between examples. It's been awhile since I've worked in Elixir, so I wasn't able to debug that problem. Instead I opted to fix this by:
1. Using distinct names in the Agents koan.
2. Combining the problematic examples sin the GenServers koan.

test/test_helper.exs
```diff
@@ -1,4 +1,5 @@
-ExUnit.start()
+timeout = 1000 # ms
+ExUnit.start(timeout: timeout)
```
iamvery, line 2: ✍️ The default timeout is 60s, so I reduced it to 1s so that tests would fail out more quickly.

iamvery (commented):
I would love other's perspective on what's going on with the process timeouts! I am open to pairing on it if my description of the issue is unclear.

lib/koans/18_genservers.ex
```diff
@@ -12,9 +12,13 @@ defmodule GenServers do
...
 
-    def start_link(init_password) do
+    def start(init_password) do
       # The __MODULE__ macro returns the current module name as an atom
-      GenServer.start_link(__MODULE__, init_password, name: __MODULE__)
+      GenServer.start(__MODULE__, init_password, name: __MODULE__)
+    end
+
+    def stop do
+      GenServer.stop(__MODULE__)
```
iamvery, line 21: ✍️ I've opted to manually manage the process instead of relying on linking. Since each koan is run in its own process, there was a race condition between the servers when running linked. Since koans are tested in order, this should clean up after itself, but even if it doesn't I think we're pretty safe as overall runtime is short.

iamvery:
✍️ I've gone ahead and self-merged since I've been unable to get review from anyone 🙃

PR #273, opened 2022-04-25
Add helpful message when processes koan assertion fails

### LINKED ISSUE #266: Match failure in 15_processes.ex - ex_unit_no_meaningful_value
Something isn't matching in Processes 173
Here's my code:

```elixir
koan "A common pattern is to include the sender in the message, so that it can reply" do
  greeter = fn ->
    receive do
      {:hello, sender} -> send(sender, :how_are_you?)
    end
  end

  pid = spawn(greeter)

  send(pid, {:hello, self()})
  assert_receive :how_are_you
end
```

And here's what I get:

```
Now meditate upon Processes
|========================>     | 173 of 209
A common pattern is to include the sender in the message, so that it can reply
Assertion failed in lib/koans/15_processes.ex:80
assert_receive(:how_are_you)

left:  :how_are_you
right: :ex_unit_no_meaningful_value
```

Interestingly, even if I change it to really forgiving match arms, it still gives me `:ex_unit_no_meaningful_value`:

```elixir
self_pid = self()
greeter = fn ->
  receive do
    _anything -> send(self_pid, :how_are_you?)
  end
end
```

I'm on macOS and Elixir version 1.12.2 with Erlang/OTP 24. Any help would be appreciated!

iamvery:
[closes #266]

# Problem
Failure was confusing in #266.

# Solution
Add message to nudge learner in the right direction.

```
Processes

Now meditate upon Processes
|                              | 0 of 36
A common pattern is to include the sender in the message, so that it can reply
Assertion failed in lib/koans/15_processes.ex:83
Sorry, I didn't get the right message. Look at the message that is sent back very closely, and try again
```

### DIFF
--- a/lib/koans/15_processes.ex
+++ b/lib/koans/15_processes.ex
@@ -77,7 +77,10 @@ defmodule Processes do
     pid = spawn(greeter)
 
     send(pid, {:hello, self()})
-    assert_receive ___
+
+    timeout = 100 # ms
+    failure_message = "Sorry, I didn't get the right message. Look at the message that is sent back very closely, and try again"
+    assert_receive ___, timeout, failure_message
   end
 
   def yelling_echo_loop do


iamvery:
@ukutaht @felipesere can you please have a quick review and confirm that this is an improvement? i just want some accountability before merging

felipesere:
LGTM

Issue #275, opened 2022-06-14
weird answer for process koan about message queue


EB-Plum:
https://github.com/elixirkoans/elixir-koans/blob/f5e3f75448e9a44bb3445d670f65ed37a216c792/lib/koans/15_processes.ex#L62-L68
```
# lib/koans/15_processes.ex. line 62
koan "Received messages are queued, first in first out" do
    send(self(), "hola!")
    send(self(), "como se llama?")
    send(self(), "1")
    send(self(), "2")
    send(self(), "3")

    assert_receive "1"
    assert_receive "3"
    assert_receive "como se llama?"
    assert_receive "hola!"
    assert_receive "2"
  end
```

i tried wrong answer(?), but somehow it passed. 
is this just my fault? or `assert_receive` acting weird?

EB-Plum:
https://hexdocs.pm/ex_unit/1.13/ExUnit.Assertions.html#assert_receive/3
according to doc, it seems `assert_receive` is not considering message order.
it only checks if message was delivered or not.

just in case,  if anyone like me want to see the queue. try code below.
```
koan "Received messages are queued, first in first out" do
    send(self(), "hola!")
    send(self(), "como se llama?")

    {:messages, queue } = Process.info(self(), :messages)

    assert Enum.at(queue, 0) == ___
    assert Enum.at(queue, 1) == ___
end
```

iamvery:
👀 #276

PR #276, opened 2022-06-21
Fix example to actually check message order as desc suggests

iamvery:
Reported in #275

# Problem
Example suggests order is important, but fails to check that in how it is written.

# Solution
Refactor example to explicitly receive messages and assert the order in which they are received.


iamvery:
@EB-Plum thanks for reporting this!

felipesere (approved):

PR #281, opened 2023-03-02
fix(ci): pin ubuntu-20 to run github actions

gpouilloux:
Fixing the issue pointed out in [this GitHub workflow execution](https://github.com/elixirkoans/elixir-koans/actions/runs/4314335021/jobs/7527274281)

We would be able to use ubuntu-22.04 when upgrading the OTP version to, at least, 24.2 (see [here](https://github.com/erlef/setup-beam#compatibility-between-operating-system-and-erlangotp))

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -10,7 +10,7 @@ jobs:
   build:
 
     name: Build and test
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
     - uses: actions/checkout@v2


iamvery:
Thank you for this fix! I've run into this on a couple other projects. I'm a little torn on just updating this project to target the latest elixir/erlang vs. pinning the os it runs in. You have any thoughts on that?

gpouilloux:
Honestly, I think both could work but it would depend on the project itself.
Personally, I would always recommend pinning to avoid such errors now and later.

You may prefer to receive [deprecation notice from GitHub](https://github.com/actions/runner-images#image-deprecation-policy) about `ubuntu-20` runner EOL instead of having your CI breaking unexpectedly.

As for upgrading to the latest OTP version, I have still little knowledge on both this project and recent OTP/Elixir changes, so that's one more argument to do it on its own PR, and _preferably_, when the CI is fixed 🙂

iamvery:
Sound reasoning. Thank you for the input, and for fixing the build 🙂

PR #282, opened 2023-07-09
Highlight that Float functions returns float

pineman:
Just a tiny improvement I felt could be made while completing these. Thank you for keeping the ruby koans tradition alive 😄

### DIFF
--- a/lib/koans/13_functions.ex
+++ b/lib/koans/13_functions.ex
@@ -25,7 +25,7 @@ defmodule Functions do
     assert first("One") == ___
   end
 
-  def repeat_again(message, times \\ 5) do
+  def repeat_again(message, times \\ 3) do
     String.duplicate(message, times)
   end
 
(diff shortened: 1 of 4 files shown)


lib/koans/13_functions.ex
```diff
@@ -25,7 +25,7 @@ defmodule Functions do
     assert first("One") == ___
   end
 
-  def repeat_again(message, times \\ 5) do
+  def repeat_again(message, times \\ 3) do
```
iamvery, line 28: ⚠️ This will cause [a test](https://github.com/elixirkoans/elixir-koans/blob/a2eaef167080a0f3223bbebb8332c00a21d0c7af/test/koans/functions_koans_test.exs#L10) to fail.

iamvery (commented):
Your changes seem pretty reasonable to me 🙂 There is one test failure that will need to be fixed. Thanks for working through these and capturing your thoughts!

lib/koans/13_functions.ex
pineman: Fixed! Sorry about that.

iamvery:
Thank you again! 🙏

PR #285, opened 2023-11-09
[in-review] Code optimisations via adding the credo

imahmedismail:
This PR involves the following changes:

- Add credo to the project and:
   - Run mix credo --all to identify possible code optimizations
   - Resolve most of the errors generated by credo such as:
     - Numbers larger than 9999 should be written with underscores: 58_127
     - Modules should have a @moduledoc tag
     - Comparison will always return true

### DIFF
--- a/test/blanks_test.exs
+++ b/test/blanks_test.exs
@@ -19,7 +19,8 @@ defmodule BlanksTest do
   end
 
   test "multiple arguments" do
-    ast = quote do: assert ___ == ___
+    # credo:disable-for-next-line Credo.Check.Warning.OperationOnSameValues
+    ast = quote do: assert(___ == ___)
 
     assert Blanks.replace(ast, [true, false]) == quote(do: assert(true == false))
   end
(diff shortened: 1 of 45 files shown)


test/blanks_test.exs
```diff
@@ -19,7 +19,7 @@ defmodule BlanksTest do
   end
 
   test "multiple arguments" do
-    ast = quote do: assert(___ == ___)
+    ast = quote do: assert true == false
```
iamvery, line 22:
```suggestion
    ast = quote do: assert ___ == ___
```

iamvery (changes requested):
This all seems pretty reasonable to me, and I don't have any strong feelings against Credo. There is one place that I think is a mistake (see requested change).

test/blanks_test.exs
imahmedismail: This is done.

iamvery:
I added configuration for credo to run in CI using `--strict` mode and fixed the suggestions it made. Thanks for the idea and getting this started!

PR #287, opened 2023-12-28
The worker/2 function from the Supervisor.Spec module is deprecated

0dayfall:
Instead of using Supervisor.Spec.worker/2, you should define your child processes within the Supervisor module using the child_spec/1 function.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Set up Elixir
       uses: erlef/setup-beam@988e02bfe678367a02564f65ca2e37726dc0268f
       with:
-        elixir-version: '1.12.3' # Define the elixir version [required]
+        elixir-version: '1.5' # Define the elixir version [required]
         otp-version: '24.1' # Define the OTP version [required]
     - name: Restore dependencies cache
       uses: actions/cache@v2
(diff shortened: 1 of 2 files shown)


iamvery (approved):
Thanks!

iamvery:
It looks like we may have to update the Elixir used for the build in order to get this passing.

iamvery (changes requested):
It looks like we may have to update the Elixir used for the build in order to get this passing.

0dayfall:
> It looks like we may have to update the Elixir used for the build in order to get this passing.

Ok, updated to Elixir v1.5

.github/workflows/elixir.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     - name: Set up Elixir
       uses: erlef/setup-beam@988e02bfe678367a02564f65ca2e37726dc0268f
       with:
-        elixir-version: '1.12.3' # Define the elixir version [required]
+        elixir-version: '1.5' # Define the elixir version [required]
```
iamvery, line 24: ❓ What this meant to be 1.15? I think the latest is 1.16.

0dayfall: Ok, I should change that

iamvery: See #288

PR #289, opened 2024-02-29
Fix inconsistent printing of meditation

Iftakharpy:
This fixes the issues I observed while solving the koans. Some of the issues I noticed are:
- Random vertical position of printing koans
- Sometimes koan intro printed before clear screen
- Next koan does not appear when all koan passes for a module

### DIFF
--- a/lib/display/progress_bar.ex
+++ b/lib/display/progress_bar.ex
@@ -4,12 +4,17 @@ defmodule Display.ProgressBar do
 
   def progress_bar(%{current: current, total: total}) do
     arrow = calculate_progress(current, total) |> build_arrow
+    progress_percentage = calculate_percentage(current, total)
 
-    "|" <> String.pad_trailing(arrow, @progress_bar_length) <> "| #{current} of #{total}"
+    "|" <> String.pad_trailing(arrow, @progress_bar_length) <> "| #{current} of #{total} -> #{progress_percentage}% complete"
   end
 
   defp calculate_progress(current, total) do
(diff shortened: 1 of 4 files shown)


lib/display/progress_bar.ex
```diff
@@ -4,14 +4,19 @@ defmodule Display.ProgressBar do
...
-    "|" <> String.pad_trailing(arrow, @progress_bar_length) <> "| #{current} of #{total}"
+    "|" <> String.pad_trailing(arrow, @progress_bar_length) <> "| #{current} of #{total} -> #{progress_percentage}% complete"
   end
 
   defp calculate_progress(current, total) do
     round(current / total * @progress_bar_length)
   end
 
+  defp calculate_percentage(current, total) do
+    Float.round(current / total * 100, 1)
```
felipesere, line 17: I wonder if this would look better without any decimal places?
```
0 of 12 -> 0% complete
```

Iftakharpy: I thought it  would be useful for people like me who wants to know the progress a little more precisely. But either way it's fine doesn't matter too much.

iamvery: I don't have a strong opinion. I tend to agree that decimals aren't terribly useful in this context.

iamvery (commented):
Changes lgtm. Since Felipe is around (and had a hand in writing this afaicr), I'll defer to him for approval.

felipesere:
Thank you for this!

PR #290, opened 2024-07-15
feat(koan): add a koan about rounding pitfalls

Goutte:
## Hint

`2.675` is actually approximated in IEE754 as: `2.67499999999999982236431605997495353221893310546875`

### DIFF
--- a/lib/koans/03_numbers.ex
+++ b/lib/koans/03_numbers.ex
@@ -109,6 +109,10 @@ defmodule Numbers do
     assert Float.round(-5.5674, 3) === ___
   end
 
+  koan "But beware of floating dragons" do
+    assert Float.round(2.675, 2) == ___
+  end
+
   koan "I want the first and last in the range" do
     first..last = Range.new(1, 10)
 


lib/koans/03_numbers.ex
```diff
@@ -109,6 +109,10 @@ defmodule Numbers do
     assert Float.round(-5.5674, 3) === ___
   end
 
+  koan "But beware of floating dragons" do
+    assert Float.round(2.675, 2) == ___
```
iamvery, line 113: There is a test suite for the koans. Would you mind adding a corresponding test for this? (My guess is that the CI run that I just started will fail and point you in the right direction)

iamvery: oof, actually it looks like CI is broken atm 😬 ...

iamvery: You should be able to run `mix test` on the app locally.

Goutte: > 49 tests, 0 failures

Thanks !  It didn't even occur to me that it was testable.  I see the Blanks now.  Pretty crafty !   :D

iamvery: Looks good! Now my turn to fix the CI 😆

PR #288, opened 2024-01-12
Update elixir version to 1.16

0dayfall:
Update elixir version to 1.16

PR #292, opened 2024-09-09
Update Elixir CI workflow

iruoy:
This fixes the failing tests and supersedes #288

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -7,25 +7,28 @@ on:
   pull_request:
     branches: [ master ]
 
+permissions:
+  contents: read
+
 jobs:
   build:
 
     name: Build and test
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     steps:
-    - uses: actions/checkout@v2
-    - name: System dependencies
+    - uses: actions/checkout@v4
+    - name: Install inotify-tools
       run: |
-        sudo apt update
-        sudo apt install -y inotify-tools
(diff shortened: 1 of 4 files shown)


.github/workflows/elixir.yml
```diff
@@ -7,25 +7,28 @@ on:
...
-        sudo apt install -y inotify-tools
+        sudo apt-get update
+        sudo apt-get install inotify-tools
     - name: Set up Elixir
-      uses: erlef/setup-beam@988e02bfe678367a02564f65ca2e37726dc0268f
+      uses: erlef/setup-beam@v1.18
       with:
-        elixir-version: '1.12.3' # Define the elixir version [required]
-        otp-version: '24.1' # Define the OTP version [required]
+        elixir-version: '1.17' # [Required] Define the Elixir version
```
iamvery, line 28: I believe this was left as a old version to ensure some backwards compatibility with running the Koans on old Elixirs. I'm not sure how important that is presently, but what is your take? Should we consider a build matrix?

iamvery (commented):
Thank you for the contribution! Would you mind adding some notes to explain the changes?

.github/workflows/elixir.yml
iamvery: As I think about this, I lean towards limiting CI to a more/most current Elixir release given that folks learning Elixir are likely to be using the latest. I'm not aware of situations where one would be limited to an old release, e.g. some computing limitation on the machine they're learning with. Thoughts?

PR #297, opened 2025-07-09
Fix Elixir Version in Dockerfile

CatalinPlesu:
It appears that on dockerhub there is no such version as 1.5

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM elixir:1.5
+FROM elixir:1.15
 RUN apt-get update && apt-get install -y inotify-tools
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM elixir:1.5
```
iamvery, line 1: Seems to be there. https://hub.docker.com/layers/library/elixir/1.5/images/sha256-d7e3150739b34ed2cd761580877b5947ea7d13232626c19d88aaeca3e8b4f2ed. I don't think we'll downgrade. If anything, we should upgrade. Thanks for the PR.

PR #299, opened 2025-08-25
Iterate on new and existing koans

iamvery:
# Problem
The examples in this project have been long-neglected. There are many core concepts that are completely missed as they didn't exist when the original set of koans were written, e.g. `with`.

# Solution
In this modern age of AI assistance, we can get a good analysis of current state and then use the LLM to build out a plan for missing examples.

NOTE: There are some additional ideas, but rather than letting this branch go one forever, I'll continue with those eventually in #301

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /deps
 erl_crash.dump
 *.ez
+.tool-versions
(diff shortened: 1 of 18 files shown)


iamvery:
@felipesere @ukutaht idk if you are still working in this space, but if so a review of these new koans would be appreciated.

iamvery:
Since I don't have any clear signal that review will be available for this, I'm going to go ahead and merge it in the hopes that someone may benefit. However, I'd _love_ feedback if anyone is willing provide that.

